Issue #2, opened 2018-03-15
Add configuration files


jackpot51:

ayoungethan:
A configuration file is ideal, as it would 
a. help user understand what the power daemon actually modifies (currently it is unclear), and 
b. allow the user to modify or create custom settings for specific use cases

For example, in my case, I need low latency audio settings for my laptop. This requires consistent CPU frequency (minimize scaling) to decrease processing latency and jitter. Both scaling and turbo boost add jitter, whereas scaling adds latency. 

However, "high performance" I believe puts on turbo boost, which adds jitter. Ideally, I would be able to create at least one "low latency" profile that lets me minimize the scaling and turn off turbo boost without losing the other performance enhances that actually help (such as higher core clock frequency).

This extension: https://github.com/konkor/cpufreq provides a front-end which is resource intensive and a bit confusing as it seems to overlap and conflict with a lot of what system76-power does, but the parameters it modifies could be extracted to a text file interface (e.g., in /etc/default/system76-power) which gives syntax and parameters for quickly and easily creating custom power profiles that can appear in the power menu.

The three default settings are fine for most users, but I could offer some language for users seeking low latency, low jitter cpu performance.

Baughn:
Another case for me: On a Dell XPS 9380, the 'balanced' power profile makes it extremely hot and leaves the fans running constantly, _if_ the laptop is connected to AC. When it's running on battery it behaves normally, so I end up using 'balanced' on battery and 'powersave' when I don't need to save power.

I'd very much like to figure out which particular setting causes this, but right now there's no way to pick and choose.

mmstick:
Turbo mode is likely the culprit. It tends to generate a lot of heat when it's activated.

Baughn:
Seems likely enough... but if that's the case, then what do I do about it?

The 'battery' profile limits CPU frequency to 50% of the base, which is 900MHz. I understand that power consumption is usually about the square of the frequency, but that still seems a bit much; there's noticeable lag. I'd like to hit some kind of Goldilocks zone.

(This laptop leans a lot on "turbo". 600MHz minimum, 1.8GHz base, 3 GHz all-core turbo, 3.6GHz peak...)

ayoungethan:
It sounds to me like its not the balanced profile so much as some
underlying process(es) running the CPU at too high speed. Balanced should
provide optimal throughput performance, battery life and thermal control in
most circumstances (except where low latency and high throughput are both
[...]

PR #13, opened 2018-05-01
Remember backlight settings for each profile

### LINKED ISSUE #6: Screen brightness changes but does not change back
When changing power profile to a less power consuming state, the screen dims in brightness. Going back to a higher performing state, the tool does not remember what was last set in that profile and keeps the screen brightness at the lower setting. 

I would expect each profile to remember the last screen brightness setting the user set before the user moved to a different power profile. When returning to the profile, the previous setting would be restored.

mmstick:
- When a profile has no backlight settings set, it will generate and set default values for that profile.
- When switching to a different profile, the current profile will remember its backlight settings.

Closes #6


brs17:
Comments from testing this pr:

This only effects screen backlight and not the keyboard backlight on galp3. On oryp3-ess keyboard backlight brightness but not color is remembered.

Does not remember settings across reboot/shutdown.

UX considerations:
I like the fact that it remembers different settings based on the power profile you are in (I see this being an improvement). But I can imagine people might want these profiles to behave differently. Thoughts @cassidyjames ?

mmstick:
Remembering settings across boots will require #2 to be implemented, which is easily done with importing [serde](https://serde.rs/) and picking whatever format we want to use (RON is interesting, but TOML's the usual goto).

brs17:
So I definitely jumped the gun on this. Having spoken with @cassidyjames, this functionality does not really align with intended ux for this feature as described here: https://github.com/pop-os/system76-power/wiki

PR #20, opened 2018-05-14
Fix for GPU not being disabled when in Intel graphics mode

jackpot51:
This will add a DBUS method and command line method to disable the NVIDIA card with bbswitch.


Makefile
```diff
@@ -38,7 +38,7 @@ vendor: .cargo/config
 	cargo vendor
 	touch vendor
 
-target/release/$(BIN):
+target/release/$(BIN): Cargo.lock Cargo.toml src/*.rs
```
jackpot51: This ensures rebuilds happen when source files are changed

debian/system76-power.service
```diff
@@ -2,6 +2,7 @@
 Description=System76 Power Daemon
 
 [Service]
+ExecStartPre=-/usr/bin/system76-power graphics power auto
```
jackpot51: This will disable the NVIDIA card when the system boots, if possible

src/main.rs
```diff
@@ -135,11 +143,10 @@ fn get_graphics() -> io::Result<&'static str> {
 }
 
 static MODPROBE_NVIDIA: &'static [u8] = br#"# Automatically generated by system76-power
-# Disabled until a fix is available: options bbswitch load_state=1
+options nvidia-drm modeset=1
```
jackpot51: We need to use modeset=1 to avoid tearing on NVIDIA graphics

PR #22, opened 2018-05-24
Remove bbswitch

jackpot51:
This is a significant rewrite of the whole switchable graphics method, to better support the `oryp4`.

Testing should be done with `nvidia-driver-396` from `ppa:dmj726/nvidia-367`


src/graphics.rs
```diff
@@ -0,0 +1,205 @@
...
+                    0x10DE => {
+                        eprintln!("{}: NVIDIA", dev.name());
+                        nvidia.push(dev);
+                    },
+                    0x8086 => {
+                        println!("{}: Intel", dev.name());
+                        intel.push(dev);
+                    },
+                    vendor => {
+                        println!("{}: Other({:X})", dev.name(), vendor);
```
mmstick: Should these last two also be `eprintln!`?

src/graphics.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    pub fn get_power(&self) -> io::Result<bool> {
+        if self.can_switch() {
+            let mut power = false;
+            for dev in self.nvidia.iter() {
+                if dev.path().exists() {
+                    power = true;
+                }
+            }
+            Ok(power)
```
mmstick: Could shorten this to:

```rust
Ok(self.nvidia.iter().any(|dev| dev.path().exists()))
```

src/graphics.rs
```diff
@@ -0,0 +1,205 @@
...
+            ))
+        }
+    }
+
+    pub fn auto_power(&self) -> io::Result<()> {
+        if self.get_vendor()? == "nvidia" {
+            self.set_power(true)
+        } else {
+            self.set_power(false)
+        }
```
mmstick: This could be

```rust
self.set_power(self.get_vendor()? == "nvidia")
```

src/client.rs
```diff
@@ -0,0 +1,178 @@
...
+        self.bus.send_with_reply_and_block(m, TIMEOUT).map_err(err_str)?;
+        Ok(())
+    }
+
+    fn auto_graphics_power(&mut self) -> Result<(), String> {
+        let m = Message::new_method_call(DBUS_NAME, DBUS_PATH, DBUS_IFACE, "AutoGraphicsPower")?;
+        self.bus.send_with_reply_and_block(m, TIMEOUT).map_err(err_str)?;
+        Ok(())
+    }
+}
```
mmstick: Could use a macro to generate these methods.

src/graphics.rs
jackpot51: Yep

src/graphics.rs
jackpot51: Yeah

src/graphics.rs
jackpot51: Yep

src/client.rs
jackpot51: I have thought about it, but I want to leave it for later. The same thing would be good for `system76-firmware`

PR #23, opened 2018-05-29
[WIP] Remove NVIDIA audio as well as NVIDIA graphics

jackpot51:
This should fix the side effects of https://github.com/pop-os/system76-dkms/pull/3


jackpot51:
I am concerned that potentially this would false detect NVIDIA chipsets, so I will likely change the logic so I can interrogate an NVIDIA device for sub-devices, and turn off those

PR #25, opened 2018-06-01
AC Event Handling

### LINKED ISSUE #8: Automatically enter High Performance when plugged in
When connected to power (and perhaps once the battery is charged to a certain percent), we should switch into High Performance mode. I don't see a strong argument for trying to save power when connected to an effectively unlimited source.

Of course we should switch back to whichever profile the user had set if they disconnect power.

### LINKED ISSUE #9: Automatically enter Battery Life when battery is low
Inversely to #8, we should consider dropping into Battery Life mode when the battery is below a certain percentage and the device is running on battery power. Like the other modes, the user should be able to toggle it back to whichever profile they'd like.

jackpot51:
- If not on battery, switch to performance
- If on battery and battery is less than 25%, switch to battery profile
- If performance was set due to being on AC and AC is unplugged, revert to prior settings
- If battery was less than 25%, and is now more than 50%, revert to prior settings
- Settings will only revert to prior settings if they were unchanged

This could use testing.

Closes #8
Closes #9


brs17:
@jackpot51 @mmstick Conflicts need to be dealt with before I can look at this. Marking WIP

mmstick:
@brs17 The conflict is fixed now.

PR #26, opened 2018-06-01
SATA Link Power Policy Control

mmstick:
- SCSI hosts may now have their power policy controlled
  - "med_power_with_dipm" will be attempted first on all levels
    - The battery profile will then attempt "min_power"
    - The balanced profile will attempt "medium_power"
    - The performance profile will attempt "max_performance"


mmstick:
Closing now that this is included in https://github.com/pop-os/system76-power/pull/31

PR #27, opened 2018-06-01
Disk Power Management

mmstick:
Support for setting:
- APM levels
- Autosuspend delay


mmstick:
Closing now that this is integrated into https://github.com/pop-os/system76-power/pull/31

PR #35, opened 2018-08-23
Set original pstate values when daemon exits

### LINKED ISSUE #34: system76-power prevents leow9-w from shutting down when a u.2 drive is connected
**Issue/Bug Description:**
Having the system76-power package installed, prevents the leow9-w from properly shutting down when a u.2 drive is connected.

**Steps I used to narrow down that system76-power is the package causing issues:**
1. Connect NVME drive as well as U.2 drive.
2. Install Ubuntu 18.04 from the Ubuntu website to the NVME drive.
3. Note that powering the machine off will work just fine.
4. Installed NVIDIA drivers from system76 repo.
5. Successfully power off the machine once again.
6. Add system76-power.
7. See that system does not successfully poweroff on it's own.

mmstick:
This may fix #34.

- When the daemon starts, the original pstate values will be stored.
- When the daemon exits by any means, either by error or by signal, those values will be restored.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,4 +1,9 @@
 [[package]]
+name = "bitflags"
+version = "1.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "dbus"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


mmstick:
@brs17 Was this after installing the package, rebooting, then shutting down?

mmstick:
May have missed a possible signal. Will add them.

brs17:
Yes @mmstick . I installed the package, rebooted, and then attempted to suspend and shutdown.

mmstick:
@brs17 Out of ideas at the moment. Perhaps logs might give some insight.

brs17:
Here is the output of `journalctl -b -1` where the previous boot failed to shutdown.

[shutdownfreeze.log](https://github.com/pop-os/system76-power/files/2318723/shutdownfreeze.log)

mmstick:
The logs are showing that this patch is doing precisely what it is supposed to do. The kernel sends SIGTERM to the daemon, and the daemon handles that signal by setting the original values before exiting. The shutdown proceeds as normal until the logs stop when `journald` itself stops.

mmstick:
Doesn't seem to be related to pstate. With the daemon disabled and manually setting the pstate values that the daemon sets on startup, it succeeds to shut down.

PR #36, opened 2018-08-24
Use clap for argument parsing

mmstick:
This will enhance the flexibility of argument parsing, and automatically handle arg parsing errors / help info generation. This could also be used to generate manpages & shell completions at compile-time.


brs17:
Per our discussion in person yesterday, maybe a few changes could be made regarding the way the options were listed (and _when_ certain ones like `--verbose` are listed). Otherwise looks good.

PR #37, opened 2018-08-30
Enable Runtime Power Management for PCI Devices

mmstick:
With this, PowerTop will report that we're setting good values for all of the reported tunables on the battery profile, except for USB autosuspend, which I don't think we want.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -12,12 +12,13 @@ use disks::{Disks, DiskPower};
 use graphics::Graphics;
 use hotplug::HotPlugDetect;
 use kbd_backlight::KeyboardBacklight;
-use kernel_parameters::{DeviceList, Dirty, KernelParameter, LaptopMode, NmiWatchdog};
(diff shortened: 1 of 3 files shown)


ebobby:
Merge!! :)

brs17:
So to my understanding, with this patch in action, I should see on the "Tunables" tab of powertop the runtime pm devices should all be labeled "Good" not "Bad".

I added the patch and do not see that as the case.

I add the patch, reboot, and then run powertop. I am doing this all on battery power. Am I missing something?

ebobby:
@brs17 are you running your deamon with the `--experimental` flag ?

brs17:
I was @ebobby.

Looks like I had to switch to "Battery Life" to see most everything go to "Good". Then switching back to "Balanced" showed more "Good"s then previously. That doesn't seem quite right. @mmstick is investigating.

brs17:
Looks good to me now. Is there any situation you might think that this could cause an issue?

brs17:
@jackpot51 maybe code review that latest commit?

ebobby:
Excellent! Ill update the AUR package asap.

PR #42, opened 2018-09-14
Configuration File, Configurable Profiles, & Custom Profiles

mmstick:
Closes #2 
Improves the situation for #38 
Requires #46 in order for the `threshold`s and `default.{battery,ac}` to have an affect.
Closes #48

### Error Handling Improvements

- Errors when setting a parameter in a profile will no longer stop the attempt to set a profile.
- Each error that occurs when setting a profile will be combined into a single error at the end.

### TOML Config

- A TOML config for the daemon is placed at `/etc/system76-power/config.toml`
- Deserialization of this config is performed by `serde`.
- Serialization is done by hand, however, to get the desired formatting, hints, and explanatory comments.
- The default settings will be used if there is an error when reading the config

### Fixes

- Logging now works for the CLI as well as the daemon.

### New Features

- A configuration script / command may now be defined for each profile
- Custom fan curve and power profiles may also be defined and set.
- New `fan-curve` subcommand to swap fan curves without restarting the daemon.
- New `config` subcommand to display and validate the config.

### Config

Example config (generated by default when the config is missing):

```toml
# This config is automatically generated by system76-power.

# The default profile that will be set on connecting to AC.
# ac = 'performance'

# The default profile that will be set on disconnecting from AC.
# battery = 'battery'

# Uncomment to enable extra untested power-saving features
# experimental = true

# [threshold]
# Defines what percentage of battery is required to set the profile to 'battery'.
# critical = 25

# Defines what percentage of battery is required to revert the critical change.
# normal = 50

# Settings for controlling fan curves
[fan_curves]
# The fan curve to set when starting the daemon. Default is 'standard'.
active = 'standard'

# The default fan curve.
# standard = [ { temp = 2000, duty = 3000 },
#              { temp = 3000, duty = 3500 },
#              { temp = 4000, duty = 4250 },
#              { temp = 5000, duty = 5250 },
#              { temp = 6500, duty = 10000 } ]

[profiles]
# The last profile that was activated.
active = 'balanced'

# [profiles.battery]
# backlight = { keyboard = 0, screen = 10 }
# laptop_mode = 2
# max_lost_work = 15
# pci = { runtime_pm = true }
# pstate = { min = 0, max = 50, turbo = false }
# radeon = { profile = 'low', dpm_state = 'battery', dpm_perf = 'low' }
# script = '$PATH'

# [profiles.balanced]
# backlight = { keyboard = 50, screen = 40 }
# laptop_mode = 0
# max_lost_work = 15
[...]

PR #46, opened 2018-09-24
AC Event Handling

### LINKED ISSUE #8: Automatically enter High Performance when plugged in

### LINKED ISSUE #9: Automatically enter Battery Life when battery is low

### LINKED ISSUE #21: Automatically enter Balanced when disconnecting from power
Specified in #8, but more explicitly stated here. When a laptop disconnects from power, we should enter Balanced mode (unless the battery is low).

### LINKED ISSUE #38: Power settings lack performance option
**NAME="Pop!_OS"
VERSION="18.04 LTS"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Pop!_OS 18.04 LTS"
VERSION_ID="18.04"
HOME_URL="https://system76.com/pop"
SUPPORT_URL="http://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=bionic
UBUNTU_CODENAME=bionic
**

**Issue/Bug Description:**
There is no way to set a performance mode. The only way to set is through the current **System76 Power extension** between:
- High Performance
- Balanced
- Battery Life

Nevertheless, it is currently **impossible** to change this extension's default to high performance when connected to electricity.

A. Settings
- Access Settings > Power
- Fail to change efficiency mode 

B. System76 Power extension
- Access extension's drop-down menu
- Change to a different value. E.G.: High performance
- Restart computer
- default goes back to Balanced

**Expected behavior:**
A. Be able to set an efficiency mode
B. Be able to set a new default

mmstick:
**Tested features**

- [x] If on AC, and the user switches to battery, the balanced profile will be set.
- [x] If on battery, and the user switches to AC, the performance profile will be set.
- [x] If on battery, and the battery drops below 25% (critical), the battery profile will be set.
- [x] If on battery, and the battery rises above 50% after previously reaching critical, the balanced profile will be set.

Closes #8
Closes #9 
Closes #21 
Closes #38 

Requires #42 to make the settings configurable


mmstick:
Finally found a solution to the build error here. We're going to need to create a compressed tarball of our vendored sources that preserves permissions outside of the schroot, so that `dpkg-buildpackage` doesn't tamper with them when it creates the source tarball. The fix to the Makefile will be pushed soon.

jackpot51:
Interesting

Makefile
```diff
@@ -35,12 +35,17 @@ update:
 	cp $< $@
 
 vendor: .cargo/config
-	cargo vendor
-	touch vendor
+	if [ ! -f vendor.tar.xz ]; then \
+		cargo vendor; \
+		touch vendor; \
+		tar pcfJ vendor.tar.xz vendor; \
+		rm vendor -r; \
```
jackpot51: I prefer `rm -r vendor`.

Makefile
```diff
@@ -35,12 +35,17 @@ update:
 	cp $< $@
 
 vendor: .cargo/config
-	cargo vendor
-	touch vendor
+	if [ ! -f vendor.tar.xz ]; then \
+		cargo vendor; \
+		touch vendor; \
```
jackpot51: touch vendor is not required

Makefile
```diff
@@ -35,12 +35,17 @@ update:
 	cp $< $@
 
 vendor: .cargo/config
```
jackpot51: the target should be vendor.tar.xz

Makefile
```diff
@@ -35,12 +35,17 @@ update:
...
-	cargo vendor
-	touch vendor
+	if [ ! -f vendor.tar.xz ]; then \
```
jackpot51: if you use the target as vendor.tar.xz, you do not need the if statement

brs17:
I am hesitant to go with these defaults...in discussing this pr in the office, we came up with a number of edge cases where this functionality is potentially not what a user would want. 

I definitely like the idea for the user to be able to set these types of settings but I am not sure these defaults are what we would want to offer out of the box.

mmstick:
Deferring additional work on this until #42 is merged.

andirsun:
@mmstick #42 is closed so I really need this PR, would be possible move forward with this PR?

dhpowrhost:
This feature would really be nice.

PR #54, opened 2018-12-01
Add method to start discrete graphics with drivers.

krieghof:
This is very handy for developers who work with tensorflow or any other software that gets a benefit of cuda.

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -147,6 +147,54 @@ impl Graphics {
         }
     }
 
+    pub fn load_nvidia(&self,) -> io::Result<()> {
+        if self.can_switch() {
+        let path = "/etc/modprobe.d/system76-power.conf";
+
+            {
+                info!("Creating {}", path);
+                let mut file = fs::OpenOptions::new()
+                    .create(true)
+                    .truncate(true)
+                    .write(true)
+                    .open(path)?;
+
+                file.write_all(MODPROBE_NVIDIA)?;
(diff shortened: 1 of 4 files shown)


src/graphics.rs
```diff
@@ -147,6 +147,54 @@ impl Graphics {
...
+        if self.can_switch() {
+        let path = "/etc/modprobe.d/system76-power.conf";
+
+            {
+                info!("Creating {}", path);
+                let mut file = fs::OpenOptions::new()
+                    .create(true)
+                    .truncate(true)
+                    .write(true)
+                    .open(path)?;
```
mmstick: This would actually be equivalent:

```rust
let mut file = File::open(path)?;
```

src/graphics.rs
```diff
@@ -147,6 +147,54 @@ impl Graphics {
...
+            self.set_power(true);
+
+            info!("Enabling nvidia module");
+            let status = process::Command::new("modprobe").arg("nvidia").status()?;
+            if ! status.success() {
+                return Err(io::Error::new(
+                    io::ErrorKind::Other,
+                    format!("modprobe: failed with {}", status)
+                ));
+            }
```
mmstick: A modprobe wrapper already exists [here](https://github.com/pop-os/system76-power/blob/master/src/modprobe.rs). You can use it like so:

```rust
modprobe::load("nvidia", &[])?;
```

PR #59, opened 2019-01-06
RHEL* support

szydell:
Modern RHEL* distros use Dracut instead of update-initramfs.
Fix to check what tool is available and uses the correct one.

+ small change to namespace scope to shorten the code.


jackpot51:
Have you confirmed that the other changes made by system76-power for graphics switching are working on RHEL systems?

szydell:
I do use it on Fedora 29. Cli switching between intel and nvidia works same as on pop.

szydell:
This pull request grows more than it should and now contains too many changes. I'll try to prepare it more granulated. ;)

PR #61, opened 2019-01-30
oryp5

jackpot51:
Add oryp5 hotplug detection


jackpot51:
This is WIP because I have not verified that it works yet with the old PCH settings.

jackpot51:
This works correctly now, would like testing on old oryp4

jackpot51:
Verified on oryp4

PR #63, opened 2019-02-07
Add GetProfile dbus method & PowerProfileSwitch dbus signal

mmstick:
Allows other applications to know which power profile is currently active. The shell extension could use this to accurately detect which profile is active, instead of default to balanced.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -48,6 +48,12 @@ impl Power for PowerClient {
         r.get1().ok_or_else(|| "return value not found".to_string())
     }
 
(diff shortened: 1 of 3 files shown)


mmstick:
Will soon be adding a signal for profile switches, so the shell extension can take advantage of it.

isantop:
Testing this works as expected on a Darter. Approval pending results on Oryx to validate graphics switching

isantop:
No breakage on Darter or Oryx. Graphics switching works as expected.

mmstick:
Great. I'll merge it then.

PR #66, opened 2019-03-11
Automatic switching of displayport mux

jackpot51:
On galp2, galp3, galp3-b, galp3-c, and darp5, this adds automatic switching of the displayport mux when no display is connected on mDP or USB-C


isantop:
Please hold off on merging this, as it was pointed out we need to do some minor additional testing

isantop:
If the computer is booted with the USB display plugged in, the display is visible in Displays, but there needs to be a mode change before the display will activate (otherwise it shuts off). The display can also be unplugged/replugged to get the same effect.

mmstick:
Is this ready for merging?

brs17:
Just updated with master @mmstick , then it's good!

PR #78, opened 2019-05-07
Support multiple I/O boards and multiple CPUs

jackpot51:
This is required to fix issues with fans on massive


jackpot51:
Only one Io board and one CPU are used for automatic fan control, leaving the second one to the default settings.

PR #88, opened 2019-05-31
Persistable power profiles

mmstick:
Enables power profiles to persist between daemon reloads and system reboots. Power profile changes will overwrite a file at `/var/lib/system76-power/power-profile`, thereby allowing the daemon to remember changes between sessions.

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -22,9 +22,10 @@ use crate::{
     Power, DBUS_IFACE, DBUS_NAME, DBUS_PATH,
 };
 
+mod persistables;
 mod profiles;
 
-use self::profiles::*;
+use self::{persistables::Persistables, profiles::*};
 
 static CONTINUE: AtomicBool = AtomicBool::new(true);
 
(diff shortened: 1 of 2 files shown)


DylanC:
@brs17 - I would also be interested in this feature. :+1:

danieleades:
not sure if you're looking for any code reviews?

`Persistables` might make more sense as an extension trait or a struct in its own right. That would make it a little easier to reuse.

as a struct-

```rust
pub struct Persist<T> where T: FromStr + ToString + Default {
    path: PathBuf,
    inner: T,
}

impl<T> Deref for Persist<T> where T: FromStr + ToString + Default, {
    type Target = T;

    fn deref(&self) -> &Self::Target {
        &self.inner
    }
}

impl<T> DerefMut for Persist<T> where T: FromStr + ToString + Default, {
    fn deref_mut(&mut self) -> &mut T {
        &mut self.inner
    }
}

impl<T> Persist<T> where T: FromStr + ToString + Default, {
    // ...
}
```

as an extension trait

```rust
pub trait Persistable<T> where T: FromStr + ToString + Default, {
    const path: &'static Path;

    // ...
   
}
```

of course you could also use an existing solution such as `confy`, which would give you persistence of any serialisable struct

PR #93, opened 2019-06-04
Support for configuration files

### LINKED ISSUE #2: Add configuration files

mmstick:
There are two locations for the configuration file:

- System Defined: `/usr/lib/system76-power/config.toml`
- User Defined: `/etc/system76-power/config.toml`

Values which are set in the user-defined configuration file are overlayed onto the system-defined configuration parameters. This means that the system config is used as a template for setting the default values, and then any values stored in the user-defined configuration file will overwrite the default values.

Maintainers are therefore required to provide a system config file with their packages. Included is the default configuration used by Pop!_OS. The default config is at `data/config.toml`.

Closes #2

### DIFF
--- a/src/daemon/config/mod.rs
+++ b/src/daemon/config/mod.rs
@@ -1,0 +1,289 @@
+pub use self::{
+    backlight::{BacklightConfig, BacklightMethod},
+    fan::FanConfig,
+    profiles::{DiskConfig, PStateConfig, ProfileConfig, ScsiPolicy},
+    radeon::{RadeonConfig, RadeonDpmState, RadeonProfile},
+};
+use std::{collections::HashMap, fs, io, path::Path};
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct Config {
+    #[serde(default)]
+    pub fans: FanConfig,
+    #[serde(default)]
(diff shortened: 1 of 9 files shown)


src/daemon/config/mod.rs
```diff
@@ -0,0 +1,289 @@
...
+                config.profiles.contains_key(required),
+                "missing {} profile in default config",
+                required
+            );
+        }
+
+        assert!(
+            config.fans.curves.contains_key("standard"),
+            "missing standard fan curve profile in default config"
+        );
```
danieleades, line 36: serde is sufficiently flexible to encode these assertions in the types. unless you're doing it at runtime in order to customise the panic message?

something along the lines of (untested)

```rust
#[derive(Clone, Debug, Deserialize)]
pub struct Config {
    #[serde(default)]
    pub fans: FanConfig,
    #[serde(default)]
    pub profiles: Profiles,
}

#[derive(Clone, Debug, Deserialize)]
struct Profiles {
   balanced: ProfileConfig,
   battery: ProfileConfig,
   performance: ProfileConfig,
   #[serde(flatten, default)]
   other: HashMap<String, ProfileConfig>,
}
```

PR #94, opened 2019-06-05
Persistable brightness

### LINKED ISSUE #82: Brightness at startup is fixed and has to be changed manually
**Hardware:**

Lenovo Thinkpad T430

 Pop!_OS 18.10
 

**Issue/Bug Description:**

Screen brightness after booting is fixed to a certain level (about 50%), no matter at what value it was set in the last session. You have to adjust the brightness manually everytime. Usually the last value should be set again.

**Other Notes:**

Others also have the problem, see: https://pop-planet.info/forums/threads/brightness-at-startup.80/

mmstick:
Fixes #82.

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -52,6 +52,7 @@ static PCI_RUNTIME_PM: AtomicBool = AtomicBool::new(false);
 fn pci_runtime_pm_support() -> bool { PCI_RUNTIME_PM.load(Ordering::SeqCst) }
 
 struct PowerDaemon {
+    initial_set:         bool,
     graphics:            Graphics,
     power_profile:       String,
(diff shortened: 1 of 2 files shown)


jackpot51:
gnome-settings-daemon already persists the backlight setting. Perhaps we can avoid overriding this when the profile is set on boot?

mmstick:
Seems to be a bug in gnome-settings-daemon that sets the slider to the last-known value, but does not actually change the brightness values, and actually sets brightness to 100% by default. At least on the oryp2-ess. Will try it out on a different model.

mmstick:
It works on the oryp5, so there's just a quirk with GNOME on the oryp2-ess that isn't setting the backlight parameters correctly.

Issue #96, opened 2019-06-19
Add profile to turn off Turbo, allowing for cooler computer


rongcuid:
This is a feature request. I request to add a profile, say _Cool_, which _turns off Turbo while not limiting frequency_. 

I know that _Battery Life_ profile already turns off Turbo, but in my case of a Serval WS 11, it also limits frequency to 2.5GHz. I request this profile because once air temperature rises above ~25 degrees Celcius, the Serval 11 cannot fun full power without overheating. The automatic throttling does not kick in fast enough to prevent hardware error, and it has crashed my computer many times. In addition, having the computer run this hot makes the keyboard too hot to use comfortably.

I have tested quite a few power settings with the intel_pstate controls under `/sys`, and notice that just by turning off Turbo, the CPU temperature may drop from a critical _99 C_ down to a much more manageable _75 C_. The CPU is still allowed to run up to 3.6GHz, instead of the 2.5GHz as limited by _Battery Life_ profile.

My current workaround is to use a SystemD service to disable turbo, but `system76-power` apparently overrides the service at random times. Thus the feature request to incorporate this new profile.

PR #97, opened 2019-06-19
Attempt to add a "Cool" profile

rongcuid:
This is my preliminary efforts in implementing the feature I requested in #96. I have never done any Rust, so the commit now is completely untested. 

Obviously I also don't have the full array of all System76 computers. I basically copied code from "balanced", registered method entry points, and simply changed "no_turbo"  to true. I suspect this is enough for a simple feature like this, but I definitely need feedback from maintainers. I will try to set up a Rust environment and test locally, too.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -47,6 +47,11 @@ impl Power for PowerClient {
         self.call_method::<bool>("Balanced", None).map(|_| ())
     }
 
+    fn cool(&mut self) -> Result<(), String> {
+        println!("setting power profile to cool");
+        self.call_method::<bool>("Cool", None).map(|_| ())
+    }
+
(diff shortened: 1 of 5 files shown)


mmstick:
You'll want to take a look at https://github.com/pop-os/system76-power/pull/93

rongcuid:
> You'll want to take a look at #93

Oh, I didn't see that. I guess I will just wait for that to be merged.

mmstick:
You can try it out now if you want. Every commit on every branch is built and packaged into their corresponding staging repositories, and there's a script in the `pop` project at `scripts/apt` which you can use to `add` a staging repository by the branch's name.

PR #100, opened 2019-07-02
Skip setting APM for md-raid devices

### LINKED ISSUE #99: failed to set disk APM level on "/dev/md0" to X: No such file or directory (os error 2)
 
 Gentoo
 
7abac874ee65d72765780d14aae9e3315c318861

**Issue/Bug Description:**

See title

* Have an md raid device node at /dev/md0
* Run the daemon (you'll get a warning but it otherwise works fine):

> [WARN] Failed to set initial profile: Errors found when setting profile:
>    \- failed to set disk power profiles: failed to set disk APM level on "/dev/md0" to 127: No such file or directory (os error 2)

* Change the profile from balanced to performance:

```
$ system76-power profile performance
setting power profile to performance
daemon returned an error message: "Errors found when setting profile:\n    - failed to set disk power profiles: failed to set disk APM level on \"/dev/md0\" to 254: No such file or directory (os error 2)"
```

**Expected behavior:**

I'm not sure what the expected behaviour should be (this could just be a misconfiguration on my part in which case feel free to close this issue) but since md  is software raid I don't think it should be trying to change anything for this "disk" and should ignore it.

**Other Notes:**

aidanharris:
Fixes #99

### DIFF
--- a/src/disks.rs
+++ b/src/disks.rs
@@ -28,7 +28,7 @@ impl Default for Disks {
         for device in blocks.flat_map(Result::ok) {
             if device.path().join("slaves").exists() {
                 if let Ok(name) = device.file_name().into_string() {
-                    if name.starts_with("loop") || name.starts_with("dm") {
+                    if name.starts_with("loop") || name.starts_with("dm") || name.starts_with("md") {
                         continue;
                     }
 


mmstick:
Thanks!

PR #101, opened 2019-07-16
Add TESTING.md

mmstick:
@brs17 Looking for advice on the formatting of the testing document. [Rendered](https://github.com/pop-os/system76-power/blob/6bf981395664365318fa8f9f2be4d153a6abbb18/TESTING.md)

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,0 +1,76 @@
+# Testing
+
+This document provides a guideline for testing and verifying the expected behaviors of the project. When a patch is ready for testing, the checklists may be copied and marked as they are proven to be working.
+
+## Checklists
+
+Tasks for a tester to verify when approving a patch.
+
+### CLI
+
+Tasks which test the behavior of the CLI client.
+
+- [ ] Power profiles can be queried and set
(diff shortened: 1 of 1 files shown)


brs17:
@mmstick, would testing the daemon and the GNOME shell extension not sufficiently test the cli?

mmstick:
@brs17 Testing either the CLI or shell extension tests the daemon.

brs17:
@mmstick, I'm asking, if testing the gnome shell extension is sufficient to cover all functionality.

mmstick:
Testing the shell extension, if it works, then you have verified that the daemon and shell extension work, but not the CLI frontend.

brs17:
I'd say then, let's move the checkboxes under `Daemon` to underneath the `GNOME Shell` header. I think that helps clarify what steps should be taken to test. Is it intended to have `GNOME Shell` be a higher header than `Daemon` and `CLI`?

I do very much like the work you've put in on this @mmstick :heart_eyes: !

mmstick:
That it is not. I will update it.

PR #111, opened 2019-10-03
Add support for PRIME hybrid graphics

crawfxrd:
NVIDIA driver 435.17 added support for PRIME render offloading. This
enables using the iGPU (Intel) as the default driver while allowing
rendering to be done on the dGPU (NVIDIA) for selected applications.
Like the other modes, a reboot is required after selecting the hybrid
mode.

    $ system76-power graphics hybrid
    $ sudo reboot

This takes advantage of gpu-manager creating the necessary Xorg config
when `/etc/prime-discrete` is set. Other distros may have their own
mechanism, or require manually writing the Xorg file. See [\[1\]](http://download.nvidia.com/XFree86/Linux-x86_64/435.17/README/primerenderoffload.html) for more
info on configuring the X server.

You can then see this in effect using `glxinfo` from `mesa-utils`:

    $ glxinfo | grep vendor
    server glx vendor string: SGI
    client glx vendor string: Mesa Project and SGI
    OpenGL vendor string: Intel Open Source Technology Center
    $ __NV_PRIME_RENDER_OFFLOAD=1 __GLX_VENDOR_LIBRARY_NAME=nvidia glxinfo | grep vendor
    server glx vendor string: NVIDIA Corporation
    client glx vendor string: NVIDIA Corporation
    OpenGL vendor string: NVIDIA Corporation

udev rules can be created [\[2\]](http://download.nvidia.com/XFree86/Linux-x86_64/435.17/README/dynamicpowermanagement.html) to allow Turing+ GPUs to be moved into a
low power state after a short time of inactivity. They are not created
here because they disable functionality the user may want.

    $ cat /sys/bus/pci/devices/<dev>/power/runtime_status
    suspended


src/graphics.rs
```diff
@@ -193,11 +230,65 @@ impl Graphics {
...
 
+    fn get_prime_discrete() -> Result<String, GraphicsDeviceError> {
+        let mut mode = String::new();
+        let mut file = fs::File::open(PRIME_DISCRETE_PATH)
+            .map_err(GraphicsDeviceError::PrimeModeRead)?;
+
+        file.read_to_string(&mut mode)
+            .map_err(GraphicsDeviceError::PrimeModeRead)?;
+
+        Ok(mode.trim().to_string())
```
mmstick: This could be shortened to

```rust
fs::read_to_string(PRIME_DISCRETE_PATH)
    .map_err(GraphicsDeviceError::PrimeModeRead)
    .map(|mode| mode.trim().to_owned())
```

src/graphics.rs
```diff
@@ -193,11 +230,65 @@ impl Graphics {
...
-                "nvidia".to_string()
+                let mode = Self::get_prime_discrete()?;
+                if mode == "on-demand" {
+                    "hybrid".to_string()
+                } else {
+                    "nvidia".to_string()
+                }
             } else {
                 "intel".to_string()
             };
```
mmstick: Given that we always return a static string, I think we could make the return value a `&'static str`, instead of allocating.

src/graphics.rs
```diff
@@ -227,6 +320,23 @@ impl Graphics {
             result.and_then(|_| file.sync_all()).map_err(GraphicsDeviceError::ModprobeFileWrite)?;
         }
```
mmstick: Perhaps it'd be better if we had this store the text to be written, and then have one call to write.

```rust
let text = if vendor == "hybrid" {
    MODPROBE_HYBRID
} else if vendor == "nvidia" {
    MODPROBE_NVIDIA
} else {
    MODPROBE_INTEL
};

file.write_all(text)
    .and_then(|_| file.sync_all())
    .map_err(GraphicsDeviceError::ModprobeFileWrite)?;
```

src/graphics.rs
```diff
@@ -227,6 +320,23 @@ impl Graphics {
...
+        let mode: &[u8] = if vendor == "hybrid" {
+            b"on-demand\n"
+        } else if vendor == "nvidia" {
+            b"on\n"
+        } else {
+            b"off\n"
+        };
+
+        info!("Setting {} to {}", PRIME_DISCRETE_PATH, String::from_utf8_lossy(mode));
+        Self::set_prime_discrete(mode)?;
```
mmstick: We can define `mode` as a `&str`, and then use `mode.as_bytes()` to get the byte representation. That would prevent the string allocation when writing to the log.

src/graphics.rs
```diff
@@ -193,11 +230,65 @@ impl Graphics {
...
+            .truncate(true)
+            .write(true)
+            .open(PRIME_DISCRETE_PATH)
+            .map_err(GraphicsDeviceError::PrimeModeWrite)?;
+
+        file.write_all(mode)
+            .and_then(|_| file.sync_all())
+            .map_err(GraphicsDeviceError::PrimeModeWrite)?;
+
+        Ok(())
```
mmstick: This could be shortened to

```rust
fs::write(PRIME_DESCRETE_PATH, mode))
    .map_err(GraphicsDeviceError::PrimeModeWrite)
```

It truncates the file if it exists, and creates it if it does not.

mmstick:
The checks will pass once https://github.com/pop-os/system76-power/pull/110/commits/d3aeeeda5340f019b0f00d1e6d5deec6334a3b35 is merged.

src/graphics.rs
crawfxrd: I wasn't sure about creating it if it doesn't exist. This file will be generated by `gpu-manager`, which does most of the work for enabling hybrid graphics ([here](https://git.launchpad.net/ubuntu/+source/ubuntu-drivers-common/tree/share/hybrid/gpu-manager.c?h=ubuntu/eoan#n1665)): generating the X11 configs for supporting PRIME, enables power manage on the PCI device (via writing to `/sys/bus/pci/devies/<dev>/power/control`), and loading the module if it's not already. All controlled by the value in `/etc/prime-discrete`.

mmstick: If we expect the file to exist when we are writing to it, and we do not want to create that file, then it may be better if we check that it exists beforehand, then either return a specific error, or allow it to be okay that nothing happened. Something like:

```rust
if !Path::new(PRIME_DISCRETE_PATH).exists() {
   return Err(PrimeModeFileMissing); // or return Ok(())
}

fs::write(...)
```

src/graphics.rs
crawfxrd: Being a trait method, I would rather that be done separately (if at all). All methods of the trait use `String` and not `&str`. I assume there is a reason for it.

mmstick: Ah, it's because allocated strings are given to the dbus daemon to send to its connected clients.

src/graphics.rs
```diff
@@ -25,6 +30,31 @@ alias nvidia-drm off
 alias nvidia-modeset off
 "#;
 
+const PRIME_DISCRETE_PATH: &str = "/etc/prime-discrete";
+
+const UDEV_NVIDIA_PM_PATH: &str = "/lib/udev/rules.d/80-nvidia-pm.rules";
+
+// http://download.nvidia.com/XFree86/Linux-x86_64/435.17/README/dynamicpowermanagement.html#AutomatedSetup803b0
+static UDEV_NVIDIA_PM_RULES: &[u8] = br#"# Automatically generated by system76-power
```
crawfxrd: I'm unsure if I should include these rules. They are "workarounds" for problems provided by Nvidia (the link above). If removed, the device may not be able to go into a low power state, but we also won't e.g. break the user's audio.

src/graphics.rs
```diff
@@ -208,6 +292,17 @@ impl Graphics {
     pub fn set_vendor(&self, vendor: &str) -> Result<(), GraphicsDeviceError> {
         self.switchable_or_fail()?;
 
+        let mode = if vendor == "hybrid" {
```
crawfxrd: I should add a check to ensure that `hybrid` can only be used on devices that support offloading.

src/graphics.rs
mmstick: I would think that it's more important for us to have working audio than to save additional power.

jackpot51:
@crawfxrd if you rebase on or merge with master, the CI should work and QA can take a look

crawfxrd:
Made a fix to only write to `/etc/prime-discrete` if PRIME is supported.

src/graphics.rs
```diff
@@ -193,11 +214,37 @@ impl Graphics {
 
     pub fn can_switch(&self) -> bool { !self.intel.is_empty() && !self.nvidia.is_empty() }
 
+    fn is_prime_supported() -> bool { path::Path::new(OFFLOADING_CONF).exists() }
```
crawfxrd: I'm thinking this will need to be exported so the gnome-shell extension can know if it should display the hybrid option (and whatever will add a context menu item to run something using the dGPU).

jackpot51: If we are relying on `ubuntu-drivers-common` to provide this information, it will need to be added to debian/control as a Depends

jackpot51: I think we could likely do this without relying on an Ubuntu specific project

mmstick: If you add a new DBus method to the daemon, you can access it in the extension by defining it [here](https://github.com/pop-os/gnome-shell-extension-system76-power/blob/master/extension.js#L29:L31), and [using it like so](https://github.com/pop-os/gnome-shell-extension-system76-power/blob/master/extension.js#L196)

crawfxrd: @jackpot51 I'll look into adding the detection without relying on `ubuntu-drivers-common` (seems to be reading values from sysfs).

crawfxrd:
Looking into it more, GPU offloading is only enabled if Intel is the primary VGA provider (`/sys/bus/pci/devices/<dev>/boot_vga`) [\[1\]](https://git.launchpad.net/ubuntu/+source/ubuntu-drivers-common/tree/share/hybrid/gpu-manager.c?h=ubuntu/eoan#n1128) and displays are connected to it (`/sys/class/drm/<card>/status`) [\[2\]](https://git.launchpad.net/ubuntu/+source/ubuntu-drivers-common/tree/share/hybrid/gpu-manager.c?h=ubuntu/eoan#n995). So in the case of laptops, we know GPU offloading is supported if it's switchable. (But what about desktops?)

It doesn't look like we can avoid `ubuntu-drivers-common`, as it is a dependency of other packages (including `pop-desktop` and `system76-driver-nvidia`). Specifically, the problem is the `gpu-manager` service provided by `ubuntu-drivers-common`. If it runs it's going to modify settings for the nvidia driver based on the value in `/etc/prime-discrete`.

isantop:
This should probably warn users on affected systems (like the `oryp4`) that Hybrid Graphics will disable external displays.

leviport:
I can enable hybrid in the terminal with `system76-power graphics hybrid` and it looks like it's working very well! The only issues I'm running into are with the gnome shell extension. Clicking on different power profiles does not change graphics modes or spawn the notification. The hybrid option also has the same text as Nvidia, saying "Enable for external displays" which seems to be incorrect. At least, I'm unable to get any external displays to work while in Hybrid mode, which I believe is expected.

Screenshot from 2019-10-22 08-43-49

leviport:
Testing on gaze14 and addw1 running Pop 18.04, it looks like the machine remains in nvidia mode after switching into hybrid mode, even though system76-power reports that it is in hybrid mode. I'm also able to use external monitors on ports that only work in nvidia mode, and Nvidia X Server Settings shows the dGPU being utilized without me launching any applications with rendering offloaded.
1804-not-hybrid

leviport:
Although I should mention that everything looks awesome in 19.10! The gnome shell extension looks and works great.

jackpot51:
@leviport I am not sure we will be able to have this feature on 18.04, so perhaps we make a branch of system76-power for eoan and only support it there for now?

lzyang2000:
So just to confirm - the feature won’t work in 1804?
Gonna buy a new laptop, deciding which version to install. Preferably 1804 because I use ROS, but the hybrid mode is very tempting.
Thanks!

crawfxrd:
Correct. You will need 19.10+.

lzyang2000:
Thanks! Looks like I will either use docker or restarting the display manager then

nrdxp:
Is this not available on the Serval WS? I just purchased a used serw11 with a gtx 1060 and a 8700 CPU and can't seem to get this to work. I was using 19.10 initially and recently upgraded to the 20.04 beta. When running `system76-power graphics hybrid` on either it returns `does not have switchable graphics`.

I am wondering if this might have something to do with the fact that this model uses a desktop CPU and so the GPU doesn't meet Nvidia's Optimus standard? I honestly have no idea if this is correct, just a hunch, but I would love to get this working as about the only thing I don't like on this machine is the dismal battery life.

crawfxrd:
This is not available on the Serval.

PR #118, opened 2019-10-24
Add notes about graphics modes in README

crawfxrd:
The hybrid mode comes with some caveats worth documenting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,47 @@
-# system76-power
-System76 Power Management
+# System76 Power Management
+
+**system76-power** is a utility for managing graphics and power profiles.
+
+## Graphics
+
+### Intel
+
+The iGPU (Intel) is used exclusively.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,2 +1,48 @@
...
+## Graphics
+
+### Intel
+
+The iGPU (Intel) is used exclusively.
+
+Lower graphical performance with a longer battery life.
+
+External monitors cannot be used in this mode. The ports are connected to the
+dGPU, which is disabled.
```
brs17: This is not always true. The gaze14 has a display that does work when in Intel mode.

brs17:
I wonder if we want to copy over the `How To` section from TESTING.md to the README as that has basic examples of how to use `system76-power`.

brs17:
This is a good start though, and much improved. Great work @crawfxrd !

Issue #57, opened 2018-12-20
Switch GPU without reboot


JeffLabonte:
**Distribution (run `cat /etc/os-release`):**

NAME="Pop!_OS"
VERSION="18.10"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Pop!_OS 18.10"
VERSION_ID="18.10"
HOME_URL="https://system76.com/pop"
SUPPORT_URL="http://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=cosmic
UBUNTU_CODENAME=cosmic 
 
**Related Application and/or Package Version (run `apt policy $PACKAGE NAME`):**

system76-power:
  Installed: 1.0.0~1545254257~18.10~d861dfd
  Candidate: 1.0.0~1545254257~18.10~d861dfd
  Version table:
 *** 1.0.0~1545254257~18.10~d861dfd 1001
       1001 http://ppa.launchpad.net/system76/pop/ubuntu cosmic/main amd64 Packages
        100 /var/lib/dpkg/status

**Issue/Bug Description:**

It would be great to be able to switch GPU without needing to reboot the PC.

There is an implementation that has been made by [MathieuGras](https://github.com/matthieugras/Prime-Ubuntu-18.04)

fliepeltje:
Alternatively, if this is not possible, it would also be great to have separate boot options for nvidia/intel in the bootmanager that you can select when powering on.

JeffLabonte:
It is possible. It is a feature that ubuntu has, also I have already used other alternatives like nvidia-xrun.

JeffLabonte:
@mmstick you should have a look on [this implementation](https://github.com/Askannz/optimus-manager).

I am currently running Arch on my laptop and I assure you that it works like a charm ( I am running a XPS 15 9560, a relatively new laptop).

mlncn:
Yes, please!  (Likewise to at least as a boot option.)

crawfxrd:
**Intel -> NVIDIA|Hybrid**:
```
system76-power graphics <hybrid|nvidia>
system76-power graphics power on
systemctl restart gpu-manager
# Logout and login
```

**NVIDIA <-> Hybrid**:
```
system76-power graphics <hybrid|nvidia>
systemctl restart gpu-manager
# Log out and log in
```

**NVIDIA|Hybrid -> Intel**
[...]

PR #119, opened 2019-10-25
Switch graphics mode without reboot

crawfxrd:
By further utilizing gpu-manager, we can perform most operations for switching graphics modes without requiring a reboot.

In order for switching and runpm to work correctly, GDM must be restarted. This will end all running GNOME sessions.

    system76-power graphics <mode>
    systemctl restart display-manager

See: #57

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -275,6 +275,22 @@ impl Graphics {
 
         const SYSTEMCTL_CMD: &str = "systemctl";
 
+        // NVIDIA driver functions are still in use, so we can't turn off power
+        // yet if switching to Intel.
+        if vendor == "nvidia" || vendor == "hybrid" {
+            self.set_power(true)?;
+        }
+
+        let status = process::Command::new(SYSTEMCTL_CMD)
+            .arg("restart")
+            .arg("gpu-manager.service")
+            .status()
+            .map_err(|why| GraphicsDeviceError::Command { cmd: SYSTEMCTL_CMD, why })?;
+
+        if !status.success() {
+            warn!("Unable to restart gpu-manager: {}", status);
+        }
+
         let action = if vendor == "nvidia" {
             info!("Enabling nvidia-fallback.service");
             "enable"
--- a/debian/control
+++ b/debian/control
@@ -19,6 +19,7 @@ Replaces: nvidia-prime
 Depends:
   dbus,
   systemd,
+  ubuntu-drivers-common,
   ${misc:Depends},
   ${shlib:Depends}
 Description: System76 Power Management


gase12:
hi whats up with the review? :)

flukejones:
Hey thanks for doing this!

leviport:
I began testing this and it's looking good, but it's definitely a WIP at the moment. Here is what I'm seeing:

- As @crawfxrd specified, switching from Intel to either Nvidia or Hybrid after a logout/login seems to work pretty well. The first switch seemed very reliable, with subsequent switches sometimes getting more error-prone.

- When switching from Hybrid to Nvidia at one point (I lost track of how many other switches preceded this switch), I logged out and back in again and saw that all gnome extensions had been disabled.

- I was able to switch from either Nvidia or Hybrid to Intel by restarting GDM with `systemctl restart display-manager`, but switching then from Intel to Hybrid/Nvidia without restarting resulted in a hang at the "Switching graphics modes" window, and that window never changed to the version with "restart now" and "restart later" buttons. 

- All of the messaging and dialogs still point to restarting after a switch, so that will all need to be updated eventually as well.

Once we're further along, we can start doing some more testing!

isantop:
I would imagine we want to have the messaging and dialog restart the session/GDM for the user. We can probably co-opt the current reboot dialog for that.

crawfxrd:
Need to investigate runpm. I'm seeing the card left `active` after switching from Intel to hybrid.

crawfxrd:
@leviport I haven't been able to reproduce the issues. I'd be interested in getting any logs for the hang if you're able to repro it.

Remaining work for #57 is:
- restart GDM
- apply run-time power management workarounds
- integrate with the shell extension

leviport:
> I'd be interested in getting any logs for the hang if you're able to repro it.
I can take another look to see if I can make it happen again, though it looks like the latest eoan build didn't succeed. Once it builds again I will see if it's still an issue and gather logs if it is.

src/graphics.rs
```diff
@@ -275,6 +275,22 @@ impl Graphics {
 
         const SYSTEMCTL_CMD: &str = "systemctl";
 
+        // NVIDIA driver functions are still in use, so we can't turn off power
+        // yet if switching to Intel.
+        if vendor == "nvidia" || vendor == "hybrid" {
+            self.set_power(true)?;
+        }
+
+        let status = process::Command::new(SYSTEMCTL_CMD)
```
crawfxrd: gpu-manager is configured to run before display-manager, so implementing a display-manager restart would allow this call to be removed.

lzyang2000:
Great feature, hope it gets merged soon.
I definitely do not want to reboot every time I switch graphics lol

brs17:
@crawfxrd are we expecting this to function in 18.04 (between Intel and NVIDIA modes)?

crawfxrd:
It looks like the required logic exists in `ubuntu-drivers-common` for bionic. I don't see why it wouldn't, but I didn't consider it.

lzyang2000:
Any plans to merge this

crawfxrd:
Because this can't be implemented cleanly (must kill all X sessions), we're going to keep requiring a reboot. Steps in #57 will still work.

PR #121, opened 2019-10-30
Only changes power profile, if not current.

jsj1027:
Queries the current power profile before attempting to change it.
If it is the requested change is the same as the current profile, the system will not attempt to switch.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -34,22 +34,25 @@ impl PowerClient {
 
         Ok(r)
     }
+
+    fn set_profile(&mut self, profile: &str) -> Result<(), String> {
+        println!("setting power profile to {}", profile);
+        self.call_method::<bool>(profile, None)?;
+        Ok(())
+    }
 }
 
 impl Power for PowerClient {
     fn performance(&mut self) -> Result<(), String> {
-        println!("setting power profile to performance");
-        self.call_method::<bool>("Performance", None).map(|_| ())
+        self.set_profile("Performance")
     }
 
     fn balanced(&mut self) -> Result<(), String> {
-        println!("setting power profile to balanced");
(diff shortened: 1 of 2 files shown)


src/client.rs
```diff
@@ -38,18 +38,36 @@ impl PowerClient {
...
-        println!("setting power profile to battery");
-        self.call_method::<bool>("Battery", None).map(|_| ())
+        let profile = self.get_profile().ok();
+        let profile = profile.as_ref().map_or("?", |s| s.as_str());
+        if profile != "Battery" {
+            println!("setting power profile to battery");
+            self.call_method::<bool>("Battery", None).map(|_| ())
+        } else {
+            Err(String::from("power profile is already set to battery"))
+        }
```
mmstick: We could condense these down to:

```rust
fn balanced(&mut self) -> Result<(), String> {
    self.profile("Balanced")
}

....

fn profile(&mut self, profile: &str) -> Result<(), String> {
    let current = self.get_profile().ok();
    let current = profile.as_ref().map_or("?", |s| s.as_str());
    if current != profile {
        println!("setting power profile to {}", profile);
        self.call_method::<bool>(profile, None).map(|_| ())
    else {
        Err(format!("power profile is already set to {}", profile))
    }
}
```

Although we may not want to return an error for a profile that's already set. Returning `Ok(())` should be fine.

jsj1027: Sounds good ill make the changes

jsj1027:
:D

jsj1027:
Is there anything else I need to do to help this commit along? Or is everything okay?

brs17:
> Is there anything else I need to do to help this pull request along? Or is everything okay?

Everything should be good from your side, currently QA has other things higher up in the queue before getting to this PR. Thank you for your patience!

PR #123, opened 2019-11-12
Blacklist i2c_nvidia_gpu in hybrid mode

crawfxrd:
Blacklist `i2c_nvidia_gpu` in hybrid mode. Prevents a possible lockup caused by this module misbehaving.


brs17:
> It required me toggling between Intel and Hybrid modes after installation, but it seems to work very well. Cool stuff!

Is that toggling required? Is that expected? I am not sure how many people would actually think to do that.

leviport:
Nothing breaks if you don't toggle, it just doesn't suspend the GPU when inactive. That's a fair point though, I'm not sure if people would think to do that. Do you think that is a blocking issue?

brs17:
> Nothing breaks if you don't toggle, it just doesn't suspend the GPU when inactive. That's a fair point though, I'm not sure if people would think to do that. Do you think that is a blocking issue?

So if we release this as is, the people who will actually see this functionality will be new installs who update before switching graphics modes and people who switch modes frequently. I think that is a very small number of users. Is there something we can do to get this in front of more users?

jackpot51:
This file needs to be removed in the postrm, similar to other files. Also, does the file break anything when installed on other systems or when not in hybrid mode?

If not, we could add it to the NVIDIA driver.

crawfxrd:
The first three rules disable the DP/HDMI audio and USB functions, which I don't want to happen in NVIDIA mode where a user may expect to use them.

As for toggling modes to enable this: setting a mode runs all the logic, even if it's already in that mode. So it should only be necessary to run `graphics hybrid` and reboot once.

brs17:
> The first three rules disable the DP/HDMI audio and USB functions, which I don't want to happen in NVIDIA mode where a user may expect to use them.
> 
> As for toggling modes to enable this: setting a mode runs all the logic, even if it's already in that mode. So it should only be necessary to run `graphics hybrid` and reboot once.

If a user who is already in hybrid graphics mode, has to re-run `graphics hybrid`, perhaps this is something that should be run automatically on boot? I do not think it expected that the the user would rerun such a command (or run it at all if switching graphics using the shell extension).

leviport:
Also I believe https://github.com/pop-os/system76-power/pull/121 made it not re-run if the current profile is selected. The user would have to switch to another mode and switch back.

brs17:
> Also I believe #121 made it not re-run if the current profile is selected. The user would have to switch to another mode and switch back.

Should we revert #121 then?

leviport:
I think #121 makes way more sense in every context except this one. Clicking on the mode you're already running shouldn't stop a user from using their machine for the ~15 seconds it takes to display the graphics mode switch dialog, only for them to click "Restart Later" and carry on with what they're doing.

crawfxrd:
#121 only applies to the power profiles. Graphics modes still run all the logic.

leviport:
:man_facepalming: You're right, sorry about that.

crawfxrd:
I'm also wondering how much work it would be to patch `ubuntu-drivers-common` to let `gpu-manager` handle it, since it's already generating the Xorg config we need.

crawfxrd:
Patching `ubuntu-drivers-common` is easy enough (although I need to learn how to make a package for it). It doesn't require writing the udev file, just writing to sysfs to remove the extra PCI functions. This method will work better for #57.

crawfxrd:
pop-os/ubuntu-drivers-common#1

flukejones:
You might be interested in the way I did this in a patch for an openSUSE build. I separated out the udev rules parts like this:

**0002-Set-UDEV-rules-required-for-powering-off-Nvidia.patch**

```
From 87891cfec727f2c69f540e26ded2871b7570bcfa Mon Sep 17 00:00:00 2001
From: Luke Jones <luke@ljones.dev>
Date: Sun, 10 Nov 2019 14:48:03 +1300
Subject: [PATCH 2/2] Set UDEV rules required for powering off Nvidia

To power off the GPU fully when using "hybrid" mode UDEV rules are
required to remove some devices which currently prevent the GPU from
powering off.
---
 src/graphics.rs | 30 ++++++++++++++++++++++++++++++
 1 file changed, 30 insertions(+)

diff --git a/src/graphics.rs b/src/graphics.rs
index 3bff527..9c021f3 100644
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -46,6 +46,15 @@ Option "PrimaryGPU" "true""#;
 static PRIMARY_GPU_END: &[u8] = br#"
 EndSection"#;
 
+const UDEV_RULES_PATH: &str = "/etc/udev/rules.d/90-system76-pm.rules";
+
+static UDEV_RULES_PM: &[u8] = br#"# Automatically generated by system76-power
+# http://download.nvidia.com/XFree86/Linux-x86_64/435.17/README/dynamicpowermanagement.html
+ACTION=="add", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x0c0330", ATTR{remove}="1"
+ACTION=="add", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x0c8000", ATTR{remove}="1"
+ACTION=="add", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x040300", ATTR{remove}="1"
+"#;
+
 const PRIME_DISCRETE_PATH: &str = "/etc/prime-discrete";
 
 #[derive(Debug, Error)]
@@ -305,6 +314,27 @@ impl Graphics {
                 [PRIMARY_GPU_BEGIN, PRIMARY_GPU_END].concat()
             };
 
+            file.write_all(&text)
+                .and_then(|_| file.sync_all())
+                .map_err(GraphicsDeviceError::ModprobeFileWrite)?;
+
+            info!("Creating {}", UDEV_RULES_PATH);
+
+            // begin section for non-separated Nvidia xorg modules
+            // eg, not put in their own directory
+            let mut file = fs::OpenOptions::new()
+                .create(true)
+                .truncate(true)
+                .write(true)
+                .open(UDEV_RULES_PATH)
+                .map_err(GraphicsDeviceError::ModprobeFileOpen)?;
+
+            let text = if vendor == "hybrid" {
+                UDEV_RULES_PM
+            } else {
+                br#"# Automatically generated by system76-power"#
+            };
+
             file.write_all(&text)
                 .and_then(|_| file.sync_all())
                 .map_err(GraphicsDeviceError::ModprobeFileWrite)?;
-- 
2.23.0
```

**90-nvidia-udev-pm-G05.rules**

```
# https://download.nvidia.com/XFree86/Linux-x86_64/435.17/README/dynamicpowermanagement.html
# Enable runtime PM for NVIDIA VGA/3D controller devices on driver bind
ACTION=="bind", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x030000", TEST=="power/control", ATTR{power/control}="auto"
ACTION=="bind", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="auto"

# Disable runtime PM for NVIDIA VGA/3D controller devices on driver unbind
ACTION=="unbind", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x030000", TEST=="power/control", ATTR{power/control}="on"
ACTION=="unbind", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="on"
```

I'm unsure of the exact benefit of doing it this way. I guess it allows the power saving udev rule to stick around - just don't know if that actually helps anything. It certainly doesn't affect running in Nvidia mode. And hybrid mode works fine.

crawfxrd:
I question the value of the bind/unbind rules at all. It's suppose to be a workaround for if the drivers are unloaded while driving a console. But in normal use you would have to restart the system to unload the drivers anyway.

As for the other rules, I think it would be better to move the logic to `ubuntu-drivers-common`, since we're already leveraging gpu-manager for creating the X11 configs and setting `power/control`. It makes most sense for it to also apply the workarounds when configuring for `on-demand` until NVIDIA fixes their drivers.

flukejones:
The way I see it is that until the required work is complete in the Linux kernel, then these rules are required when in hybrid mode only.

The following rules are perfectly okay to have enabled all the time regardless of intel/hybrid/nvidia.

```
ACTION=="bind", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x030000", TEST=="power/control", ATTR{power/control}="auto"
ACTION=="bind", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="auto"
```

The patch I posted for `system76-power` only applies the work-around rules when `hybrid` mode is selected, this prevents the issue of having sound/usb drivers disabled in other modes. No laptop I'm aware of outputs through the hdmi/dp ports when in Intel==primary  mode as they are hardwired to the Nvidia card which means that these drivers aren't going to be used anyway.

I've quite honestly not had any issues doing it this way in openSUSE. Unsure of the cause of crash/freeze when using Pop!_OS with this.

src/graphics.rs
```diff
@@ -56,6 +78,8 @@ pub enum GraphicsDeviceError {
     Remove { device: String, why: io::Error },
     #[error(display = "failed to rescan PCI bus: {}", _0)]
     Rescan(io::Error),
+    #[error(display = "filed to modify to NVIDIA udev rules: {}", _0)]
```
crawfxrd: Fixed this typo, but don't have it pushed.

crawfxrd:
On `linux-image-5.3.0-7625-generic` (5.3.0-7625.27\~1576337002\~19.10\~bc3488b) and `nvidia-driver-440` (440.31-1pop1\~1573485407~19.10\~af08493), we're seeing the card suspend without applying any workarounds (so just using `NVreg_DynamicPowerManagement`).

flukejones:
@crawfxrd is that also true for the 5.4.x kernel? I can't use the 5.3.x kernels because they have borked FFB for my steering wheel (I play driving sims a lot).

crawfxrd:
I'm not seeing this behavior on `linux-image-unsigned-5.4.4-050404-generic` (5.4.4-050404.201912172055) from the [kernel PPA](https://kernel.ubuntu.com/~kernel-ppa/mainline/v5.4.4/). The card is left `active`.

hugglesfox:
I also cannot reproduce the results on `linux-image-5.3.0-7625-generic`. Without any udev rules, my card stays active.

crawfxrd:
I just tried with a clean install to confirm. Even without this change, on linux-image-5.3.0-7625-generic I'm seeing

```
$ cat /sys/bus/pci/devices/0000\:01\:00.0/power/runtime_status
suspended
```

On linux-image-5.3.0-20-generic and linux-image-5.3.0-23-generic I see it remain `active`.

flukejones:
Changes in this branch cause hard-lock if using hybrid mode.

The cause is `i2c_nvidia_gpu` kernel module, and this needs to be blacklisted in hybrid mode at least.

I'm getting a tingling sense of deja-vu about this for some reason, feels like something that was covered before but perhaps under different circumstances.

flukejones:
@crawfxrd these are the changes I made to `graphics.rs`:

```
static MODPROBE_NVIDIA: &[u8] = br#"# Automatically generated by system76-power
options nvidia-drm modeset=1
"#;

static MODPROBE_HYBRID: &[u8] = br#"# Automatically generated by system76-power
# http://download.nvidia.com/XFree86/Linux-x86_64/435.17/README/dynamicpowermanagement.html
options nvidia NVreg_DynamicPowerManagement=0x02
blacklist i2c_nvidia_gpu
"#;
```

Working without flaw for days now.

crawfxrd:
I'm convinced the runtime pm issues are fixed in the kernel we use.

At the very least, the issues with the audio device are fixed in Ubuntu's kernel [\[1\]](https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1840835).

crawfxrd:
Ubuntu 20.04 also does not need the udev rules

```
system76@oryp5:~$ apt-cache policy linux-image-$(uname -r)
linux-image-5.4.0-26-generic:
  Installed: 5.4.0-26.30
  Candidate: 5.4.0-26.30
  Version table:
 *** 5.4.0-26.30 500
        500 http://us.archive.ubuntu.com/ubuntu focal/main amd64 Packages
        100 /var/lib/dpkg/status

system76@oryp5:~$ lspci | grep NVIDIA
01:00.0 VGA compatible controller: NVIDIA Corporation TU106M [GeForce RTX 2070 Mobile] (rev a1)
01:00.1 Audio device: NVIDIA Corporation TU106 High Definition Audio Controller (rev a1)
01:00.2 USB controller: NVIDIA Corporation TU106 USB 3.1 Host Controller (rev a1)
01:00.3 Serial bus controller [0c80]: NVIDIA Corporation TU106 USB Type-C UCSI Controller (rev a1)

system76@oryp5:~$ cat /sys/bus/pci/devices/0000\:01\:00.*/power/runtime_status
suspended
suspended
suspended
suspended
```

flukejones:
For completeness, I'm using the 5.6.6 kernel because it contains support for a fan controller I have, and the udev rules are not required for that kernel either. I do recall that various patches to do with this were being upstreamed, but I'm unaware of which kernel version they made it in to. Ubuntu is perhaps backporting them.

```
[/h…/luke] (-) 
[luke@zephyrus]$ apt-cache policy linux-image-$(uname -r)
linux-image-5.6.6-050606-generic:
  Installed: (none)
  Candidate: (none)
  Version table:

[/h…/luke] (-) 
[luke@zephyrus]$ system76-power graphics
hybrid

[/h…/luke] (-) 
[luke@zephyrus]$ lspci | grep NVIDIA
01:00.0 VGA compatible controller: NVIDIA Corporation TU106BM [GeForce RTX 2070 Mobile] (rev a1)
01:00.1 Audio device: NVIDIA Corporation TU106 High Definition Audio Controller (rev a1)
01:00.2 USB controller: NVIDIA Corporation TU106 USB 3.1 Host Controller (rev a1)
01:00.3 Serial bus controller [0c80]: NVIDIA Corporation TU106 USB Type-C UCSI Controller (rev a1)

[/h…/luke] (-) 
[luke@zephyrus]$ cat /sys/bus/pci/devices/0000\:01\:00.*/power/runtime_status
suspended
suspended
suspended
suspended
```

PR #124, opened 2019-11-14
Fancurves

WatchMkr:
Adds a separate Core X (Thelio Major B) fan curve and modifies the standard desktop fan curve. The new standard (Thelio) fan curve better leverages the CPU heat sink and large intake fan to reduce fan noise by 7 decibels for most CPU work loads.


WatchMkr:
@crawfxrd please make the necessary updates.

PR #130, opened 2019-11-22
Log caught signals as errors

crawfxrd:
These are not part of normal process flow and indicate something has gone wrong. Log them higher to make them easier to identify.


mmstick:
We should definitely move that log statement away from the handler and into the daemon's main loop.

PR #134, opened 2019-12-19
Upgrade to dbus 0.7.1

crawfxrd:
Ref: https://github.com/diwic/dbus-rs/blob/master/dbus/changes-in-0.7.md


mmstick:
We should be able to remove all those trait objects and replace them with impl Trait returns. Remnants of an era when impl Trait didn't exist.

Issue #135, opened 2019-12-24
Backport switching graphics without reboot to bionic


brs17:
Backport https://github.com/pop-os/system76-power/pull/119/ to master_bionic if possible.

lzyang2000:
If I want to test it in 1804 - how do I build the package?
Thanks!

mmstick:
Within the pop-os/pop repository, use the scripts/apt script to add the staging branch to your system.

crawfxrd:
The behavior of GDM/gnome-shell after restarting `display-manager` is too unreliable to introduce this workaround in 18.04.

PR #138, opened 2019-12-26
bionic: Switch graphics mode without reboot

crawfxrd:
See: #119
Resolves: #135

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -19,6 +19,7 @@ Replaces: nvidia-prime
 Depends:
   dbus,
   systemd,
+  ubuntu-drivers-common,
   ${misc:Depends},
   ${shlib:Depends}
 Description: System76 Power Management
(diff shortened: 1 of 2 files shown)


crawfxrd:
On 18.04, attempting to restart GDM while still logged in results in the system becoming unstable. I am left in the user session and logout/login no longer works.

```
[  887.583680] gnome-session-c[7859]: segfault at 0 ip 0000000000000000 sp 00007ffcd9e1be18 error 14 in gnome-session-check-accelerated[555a80c4a000+2000]
[  887.583683] Code: Bad RIP value.
```

Logging out of all sessions and then restarting GDM seems to work. But this is even more inconvenient. Now you need another login (another vt or ssh) to restart GDM.

But this won't work right now. I will also need to partially backport #116.

crawfxrd:
The behavior of GDM/gnome-shell after restarting `display-manager` is too unreliable to introduce this workaround in 18.04.

PR #150, opened 2020-04-24
0.1.4 - add addw2 and gaze15

jackpot51:
Rebase and tag when approved


brs17:
Looks good on 1660ti gaze15.

PR #160, opened 2020-05-31
Update to Rust 1.41.0

jackpot51:


jacobgkau:
No regressions on gaze14. All graphics switching and power profile functionality is working as expected. Discovered a non-regression issue while testing, opened #172 for that.

Profile switching also works as expected on darp6. @leviport is still going to do some testing with hardware he's in possession of before QA approves.

Issue #156, opened 2020-05-23
Desktop AMD/Nvidia Dual GPU breaks SteamVR


robobenklein:
**Distribution (run `cat /etc/os-release`):**
 
 Pop!_OS 20.04
 
**Related Application and/or Package Version (run `apt policy $PACKAGE NAME`):**

SteamVR beta 1.12.4

**Issue/Bug Description:**

I can no longer run SteamVR and make use of a secondary Nvidia GPU at the same time. (Not even for CUDA or headless workloads.)

When the graphics mode is set to `hybrid` (so that the desktop still runs on my primary AMD GPU) SteamVR fails to launch. If I set it to `nvidia` all the desktop rendering goes to the Nvidia GPU and my overall performance suffers and I can't make use of my primary GPU.

When I set the mode to `integrated` the primary AMD card is used for everything and SteamVR works, however I can no longer make use of the Nvidia card because the driver is unloaded.

**Steps to reproduce (if you know):**

Attempt to launch SteamVR while under different GPU switching modes with two different GPUs.

**Expected behavior:**

Either `hybrid` mode should not cause SteamVR to fail or `integrated` mode needs to be changed or extended as a new mode to allow usage of the secondary card as a headless compute unit.

**Other Notes:**

System:
AMD Ryzen 2700
Primary GPU: Vega 56
Secondary GPU: GTX980
VR Headset: HTC Vive

Running unmodified Pop Shell, no 3rd party (non-System76) drivers installed.

I realize the intent here is mainly for  laptop GPU switching, but having system76-power installed should not have caused a regression in desktop behaviour. Previously I was able to make use of the AMD card as a primary GPU normally and only use the Nvidia GPU for CUDA.

crawfxrd:
> SteamVR fails to launch

Do you have logs?

robobenklein:
Switch mode to hybrid and reboot, start steam, then vr:

![image](https://user-images.githubusercontent.com/5423266/82741830-5cd3f680-9d24-11ea-90e2-005416025f7d.png)

Error code is 307, I can't immediately point out any specific error in the logs that's obvious. (log files https://gist.github.com/robobenklein/f03af509608552f70c828c84fee94070 )

Perhaps I need to debug the vrcompositor binary? My suspicion is that it's getting some kind of information about available display units and incorrectly choosing the "dedicated" graphics.

crawfxrd:
Can you also provide the output of `/var/log/gpu-manager.log` while in hybrid mode?

[...]

PR #162, opened 2020-06-03
Add "compute" graphics mode

crawfxrd:
Introduce a new "compute" mode that performs all rendering on the integrated graphics, but makes the dGPU available on the bus (with the NVIDIA drivers loaded) as a compute node.

Replaces: #54
Ref: #156


crawfxrd:
Would it be better to just add a `--compute` argument to `integrated` mode?

jackpot51:
@crawfxrd I like having it be a separate mode

jacobgkau:
With system76-power 1.1.5\~1591814391\~20.04\~baa0eed, I can switch to compute mode using the CLI, but compute mode is not listed in the top-right GNOME drop-down, and when compute mode is active, the dropdown doesn't show a dot next to any of the options. Are there plans to update gnome-shell-extension-system76-power to add compute mode to the graphical list, or is this the intended experience?

crawfxrd:
Just pushed the extension changes: pop-os/gnome-shell-extension-system76-power#45

crawfxrd:
I would have to defer to upstream GNOME devs on how to handle that. We already patch GNOME to not show the launch item in NVIDIA-only mode. Having it not show in compute mode would require inspecting the current Xorg configuration to see if `AllowNVIDIAGPUScreens` is set.

> Not sure if this one's an issue with Blender or compute mode.

It's compute mode; it looks like I will have to explicitly load the NVIDIA modules in the daemon so apps can use it.

crawfxrd:
I've fixed the issue with turning the GPU on/off in compute mode  by overriding `modeset` to off.

Still need to figure out why gpu-manager unloads the drivers after us (even with `Wants=gpu-manager.service`). Otherwise, user has to modprobe after logging in.

PR #169, opened 2020-06-08
Add oryp6

crawfxrd:


jacobgkau:
On oryp6 15" in Integrated Graphics mode, plugging in an external display correctly shows the "Switch to NVIDIA to use external displays" dialog, and clicking "Switch to NVIDIA" correctly initiates a graphics switch.

Clicking "Restart and Switch" currently seems to power off rather than reboot. This may be related to the power off/reboot issues in https://github.com/system76/firmware/pull/218.

PR #175, opened 2020-07-17
fix(TESTING.md&&Makefile): The cmd about 'graphics power state' in TESTING.md is wrong.Fix the problem of missing .service file after make install.

KT-lcz:
* The cmd about 'graphics power state' in TESTING.md is wrong.
  * Query discrete graphics power state: system76-power graphics power
  * Set discrete graphics power state:  system76-power graphics power [ on | off]
* In Makefile, I suggest that the cmd about the .service file should be added in make [install | uninstall]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,10 +35,12 @@ distclean:
 install: all
 	install -D -m 04755 "target/release/$(BIN)" "$(DESTDIR)$(bindir)/$(BIN)"
 	install -D -m 0644 "data/$(BIN).conf" "$(DESTDIR)$(sysconfdir)/dbus-1/system.d/$(BIN).conf"
+	install -D -m 0644 "debian/$(BIN).service" "$(DESTDIR)$(sysconfdir)/systemd/system/$(BIN).service"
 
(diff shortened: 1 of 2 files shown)


KT-lcz:
> Thank you for correcting those typos! I see that you also made an unrelated change-- it's typically best to mention all of the changes you're making in the PR description.
> 
> The second change just installs/uninstalls the systemd unit file for the system76-power service. The service isn't enabled by default like it is when you install the package (even though `vendor preset` still shows as `enabled`), but including the unit file so the user can enable the service themselves seems better than not including the unit file, so the change looks good to me.
> 
> (If a user installs with the `make install` script, enables the service, then uninstalls with `make uninstall`, then they will be left with a broken link from `/etc/systemd/system/multi-user.target.wants/system76-power.service` to `/etc/systemd/system/system76-power.service`; but that doesn't seem to impact the system, and the broken link can be removed with `systemctl disable system76-power`.)

Thank you for your remind, this is my negligence.I will  update PR description.

PR #177, opened 2020-08-04
feat: Methods for getting/setting keyboard backlight colors

ids1024:
This should help with implementing https://github.com/pop-os/gnome-control-center/issues/102.

So far I haven't updated the CLI to support this. I'm not sure exactly what interface we would want there.

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -4,6 +4,8 @@ use dbus::{
 };
 use std::{
     cell::RefCell,
+    collections::HashMap,
+    fs,
     rc::Rc,
     sync::{
         atomic::{AtomicBool, Ordering},
@@ -25,6 +27,8 @@ use crate::{
 mod profiles;
 
 use self::profiles::*;
+
+const KBD_BACKLIGHT_DIR: &str = "/sys/class/leds/system76::kbd_backlight";
 
 static CONTINUE: AtomicBool = AtomicBool::new(true);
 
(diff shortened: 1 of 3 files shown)


src/daemon/mod.rs
```diff
@@ -146,6 +150,31 @@ impl Power for PowerDaemon {
     fn auto_graphics_power(&mut self) -> Result<(), String> {
         self.graphics.auto_power().map_err(err_str)
     }
+
+    fn get_keyboard_colors(&mut self) -> Result<HashMap<String, String>, String> {
+        let mut colors = HashMap::new();
+        for entry in fs::read_dir(KBD_BACKLIGHT_DIR).map_err(err_str)? {
+            let entry = entry.map_err(err_str)?;
+            if let Some(name) = entry.file_name().to_str() {
+                if name.starts_with("color_") {
```
jackpot51, line 159: This won't match System76 open firmware which has a file called color. Remove the underscore and it should be good

ids1024: It also needs to change the code removing/adding `color_` in the keys when reading/writing. I just made it use the full filename including `color_` as the keys in the HashMap.

src/daemon/mod.rs
```diff
@@ -26,6 +28,8 @@ mod profiles;
 
 use self::profiles::*;
 
+const KBD_BACKLIGHT_DIR: &str = "/sys/class/leds/system76::kbd_backlight";
```
jackpot51, line 31: In open firmware systems, this is `/sys/class/leds/system76_acpi::kbd_backlight`. The most generic thing is to search /sys/class/leds for anything ending in `:kbd_backlight`. There can be something in between the two colons, which is why I only have one there.

ids1024: > The most generic thing is to search /sys/class/leds for anything ending in `:kbd_backlight`.

Hm. That complicates things if there can also be multiple drivers providing keyboard leds, which may contain files with the same name starting in `color`.

Is this `/sys/class/leds/*:*:kbd_backlight/color*` interface common to various Linux drivers, or is it just how the System76 drivers are doing backlight colors?

jackpot51: You could test for starts with "system76" and ends with "kbd_backlight" to be more specific, but catch both variants

jackpot51: I imagine launch will show up as "system76_launch::kbd_backlight"

ids1024: > I imagine launch will show up as "system76_launch::kbd_backlight"

I suppose that's another consideration for designing both this API and the user interface for https://github.com/pop-os/gnome-control-center/issues/102: a user is quite likely to have their built-in keyboard and the launch keyboard. Do we want to have separate color controls for both keyboards? We'd probably want to show a useful name for each (probably not displayed as `system76_acpi`), etc.

ids1024:
If we want to only expose a single backlight color for the keyboard, perhaps it would be best to only expose this as a single `keyboard_color` property. Plus, probably a second property specifying whether or not a recognized keyboard is connected.

* On startup, `system76-power` can read from the device to see what the current color is, or set some default.
* When the color is changed, the daemon sets every `/sys/class/leds/system76_*:*:kbd_backlight/color*` file to the same value.
* When the fn key combinations are used to set the backlight (which I believe is handled by the EC?), in one way or another system76-power needs to detect this.
* It should also detect if the Launch keyboard is plugged in, to set the color/brightness.

Exactly how this works will depend on the precise behavior we want, which is being discussed in https://github.com/pop-os/gnome-control-center/issues/102.

ids1024:
Closing. I'll work on the API on the `pop-keyboard` branch so there can be packages for this as well as the applications using the API. A new PR can be opened when this in a more satisfactory form.

PR #180, opened 2020-08-11
Bonw14 keyboard backlight

jackpot51:
Will sync color and brightness to the HID LED controllers. Both the lightguide and keyboard

Rebase and tag when approved

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5,7 +5,7 @@ name = "ansi_term"
 version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
+ "winapi 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


jackpot51:
Suspend/resume is not working. This conflicts with the boot animation

jacobgkau:
I've got this locally built on the bonw14, looks like something needs to be done for CI to try building it again.

Keyboard backlight control does work with this change. Keyboard color is not saved yet (I know we're working on that elsewhere) but brightness is saved, like other Open Firmware machines. I do notice that the keyboard is still on full brightness with the default blue at the cryptsetup screen, which is not the case for other Open Firmware machines.

One other thing I notice is that the lightguides could use some tweaking (not sure what's possible with those.) I see that the two back LED points attempt to follow the keyboard color. The front right LED point changes with the keyboard color but does not match the keyboard color, and the front left LED point seems to always be red.

Suspend/resume seems to be working for me right now. On the fourth suspend/resume, the boot animation (color cycle) continued playing a few seconds after I'd logged back in, then the color reverted to blue (which I did not have it set to when I suspended), although I was able to change the color again after that.

jackpot51:
In the latest push, I just increased the version to 1.1.7

jackpot51:
Thanks. I force pushed again. This time to remove the 10 second timeout which I worry would introduce wear on the EC flash, if it is using that to save the setting.

PR #182, opened 2020-08-18
hotplug: Add function to get Realtek subsystem device

crawfxrd:
Use the subsystem device ID of the Realtek device instead of the host bridge to identify board variants.


jackpot51:
A warning - this will not be set correctly for variants yet. I think the NVIDIA card is the safer thing to test.

PR #183, opened 2020-08-18
gaze15: Determine variant by NVIDIA device

crawfxrd:
Variants are not currently used in coreboot for gaze15, so all models have the same subsystem device ID. Use the ID of the NVIDIA device to determine the variant instead.


jackpot51:
PCI IDs:
- 1650: 10de:1f99
- 1650 Ti: 10de:1f95
- 1660 Ti: 10de:2191

PR #185, opened 2020-08-25
improv: Use anyhow::Error instead of String for error types

ids1024:
This is generally a better practice. It seems to make the code slightly neater, or comparable.

BREAKING CHANGE: This changes the public API of the crate. However, it shouldn't change the behavior of the program other than somewhat different error messages for crashes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,6 +9,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.32"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
 name = "atty"
 version = "0.2.14"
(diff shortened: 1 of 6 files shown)


jacobgkau:
The error messages for an unexpected argument and for the daemon needing to be run as root look identical before and after the change. The error message for if the daemon's not running is less descriptive after the change. Before:

```
daemon returned an error message: "The name com.system76.PowerDaemon was not provided by any .service files"
```

After:

```
daemon returned an error message
```

The former message gives a bigger hint to check if the system76-power service is running. Is removing that part necessary/intentional?

ids1024:
Hm. No, not intentional. That's definitely not a descriptive enough message.

bflanagin (commented):
This PR seems to be out of date, and needs attention. Will reassess when updated.

PR #190, opened 2020-09-04
com.system76.PowerDaemon.Keyboard dbus API

ids1024:
The [`keyboard-backlight`](https://github.com/pop-os/keyboard-backlight) repository has code using this. The API can be used with the code there to add a brightness/color setting in Gnome Control Center, and can later be used for the Keyboard Configurator.

This is rather more complicated than my previous attempt at an API for this (https://github.com/pop-os/system76-power/pull/177), but the general design is hopefully fairly robust and can be extended for more functionality we may need.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -1,11 +1,12 @@
 use crate::{err_str, Power, DBUS_IFACE, DBUS_NAME, DBUS_PATH};
 use clap::ArgMatches;
-use dbus::{arg::Append, ffidisp::Connection, Message};
+use dbus::{arg::Append, blocking::{BlockingSender, Connection}, Message};
 use pstate::PState;
 use std::io;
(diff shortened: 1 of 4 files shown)


ids1024:
This doesn't quite expose all the functionality needed for the design in https://github.com/pop-os/gnome-control-center/issues/102. That will also require the EC to persist settings, handle the color change key (instead of the kernel driver handling it), and support changing the list of colors that are switched better.

And of course the Keyboard Configurator will need a few more things, mainly support for changing the keymap.

leviport (approved):
I can't figure out how to test the keyboard-backlight app, but this PR causes no regressions, so I'm satisfied with it.

ids1024:
@jackpot51 Thoughts on this?

PR #191, opened 2020-09-16
Remove run-time power management setting

crawfxrd:
ubuntu-drivers-common 0.8.6 will handle setting runtimepm itself, based on the supported-gpus.json file provided by the 450.57+ drivers.


jackpot51 (approved):

jacobgkau:
This has been a little tricky to test. Using `/sys/bus/pci/devices/.../power/runtime_status` as an indication of the GPU's power state, this seems to be working on Groovy. On addw2 (2070), the GPU is suspended when not in use, becomes active when used, and goes back to suspended when no longer being used (although it seems to take longer to suspend after boot after the change.)

Galp5 (1650) has never worked properly yet in Hybrid or Compute mode, and it looks like gaze15 (1650) is having some problems with Compute even without this change (the GPU sometimes doesn't come out of suspend when I try to use it and I get `Unable to determine the device handle` when running `nvidia-smi`, similar to what galp5 is currently doing.)

On gaze14 (1650), the GPU doesn't suspend in hybrid mode after booting after this change, where it did suspend after booting before the change. This machine's running Focal, and Focal is still on `ubuntu-drivers-common` 0.8.4. Since `ubuntu-drivers-common` didn't get `supported-gpus.json` support until 0.8.6, do we want to update our Focal package to 0.8.6 or create a `master_focal` branch of system76-power so as not to remove runtimepm support from Focal machines?

crawfxrd:
I suppose we can do without this, assuming NVIDIA drivers just print a warning on an unsupported parameter being set.

It wouldn't be much effort to package 0.8.6.1 for 20.04 either.

bflanagin:
So has this PR come to the end of the line? Or is there still more to do and consider?

crawfxrd:
We can wait until we stop building for focal for this change.

(Or when system76-power implements everything gpu-manager does.)

crawfxrd:
Upstream has released 0.8.6.5 for focal.

PR #193, opened 2020-09-29
Disable loading nvidia-drm in compute mode

crawfxrd:
Prevent the GPU from being used to render displays while in compute mode, which we see happen with the new OutputSink feature.

Fixes: 6c354b745eac6 ("Add "compute" graphics mode")


jacobgkau (approved):
I've tried testing this on the gaze15 and the addw2.

On the current NVIDIA driver (440.100):

- Plugging in an external display while in Compute mode prompted to reboot into NVIDIA mode, and the display did not start working.
- Rebooting into Compute mode with an external display plugged in resulted in the external display being run off of the GPU.

After the change:

- Plugging in an external display while in Compute mode still prompts to reboot into NVIDIA mode, and the display still doesn't start working.
- Rebooting into Compute mode with an external display plugged in results in GDM failing to load on the addw2, but works as expected (external display ignored) on the gaze15.

On NVIDIA 455.23:

- Plugging in an external display while in Compute mode prompted to reboot into NVIDIA mode, but the display started working anyway.
- Rebooting into Compute mode with an external display plugged in resulted in the external display being run off of the NVIDIA GPU.

After the change:

- Plugging in an external display while in Compute mode still prompts to reboot into NVIDIA mode, and the display doesn't start working.
- Rebooting into Compute mode with an external display plugged in results in GDM failing to load on the addw2, but works as expected (external display ignored) on the gaze15.

To summarize, while the gaze15 successfully ignores external displays in Compute mode, the addw2 requires physically unplugging the external displays in order to load GDM in Compute mode. The addw2 also fails to load GDM in Integrated mode if an external display is plugged in, so this behavior is at least consistent. This change accomplishes the goal of not allowing external displays to be run from the NVIDIA GPU in Compute mode.

jackpot51 (approved):

PR #194, opened 2020-10-14
Move XEON curve max fans from 83 to 78

WatchMkr:


jackpot51 (approved):

jackpot51:
@pop-os/quality-assurance I'm assured by @WatchMkr this was tested on massive and did better than the current curve, so I am merging

PR #195, opened 2020-10-19
Use dbus 0.9, dbus::blocking, dbus-crossroads, and dbus-tokio

ids1024:
`dbus::ffidisp` is deprecated in favor of `dbus::blocking`, and with the latest release, `dbus::tree` has been moved to a separate crate, and `dbus-crossroads` is meant to replace it. This updates each of these things to the non-deprecated versions.

This also uses dbus-tokio, which makes it possible to have methods that are implemented asynchronously. My immediate goal was to have a sane way to invoke polkit for setting charge thresholds (without blocking the main thread for too long), but this should generally offer more flexibility.

It may be desirable to reimplement all the `Power` methods to be `async`, but this doesn't currently do that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,8 +77,8 @@ dependencies = [
 ]
 
 [[package]]
-name = "dbus-tree"
-version = "0.9.0"
+name = "dbus-crossroads"
+version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


jackpot51 (approved):

jacobgkau (approved):
Still able to query and set graphics modes and power profiles, on Focal and Groovy, using the CLI and GUI.

PR #197, opened 2020-10-26
feat: Add dbus API and CLI for battery charge thresholds

ids1024:
The dbus API authenticates with Polkit, which should prompt for an administrator password, or run without prompt when the client in root.

Thresholds can be displayed in the CLI with `system76-power charge-thresholds`, and set with `system76-power charge-thresholds <start> <end>`.

This is currently limited to System76 hardware with open firmware, to avoid worrying for now about things like systems with multiple batteries, only having an end thresholds, and other differences between implementations.

NOTE: I haven't tested this fully myself since the firmware I'm currently running doesn't have charge thresholds.


leviport:
I'm seeing something interesting if I mix up start and end numbers:
```
system76@pop-os:~$ system76-power charge-thresholds 70 90
system76@pop-os:~$ system76-power charge-thresholds
Start: 70
End: 90
system76@pop-os:~$ system76-power charge-thresholds 85 60
system76@pop-os:~$ system76-power charge-thresholds
Start: 85
End: 90
system76@pop-os:~$ system76-power charge-thresholds 60 70
system76@pop-os:~$ system76-power charge-thresholds
Start: 60
End: 70
system76@pop-os:~$ system76-power charge-thresholds 90 80
system76@pop-os:~$ system76-power charge-thresholds
Start: 60
End: 80
system76@pop-os:~$

```
Some situations lead to only one endpoint being set. Would it be more user-friendly to show an error if the first number is larger than the second? Especially since the levels aren't displayed after they are set, so what happens might not be what the user expected to happen and they wouldn't know about it unless they ran `system76-power charge-thresholds` afterwards to check.

ids1024:
Interesting. I suppose that is the behavior of the EC or the kernel. The CLI interface doesn't have to be *too* user friendly, but it's probably reasonable to added a check here.

I've amended my commit to error when `end < start`.

src/charge_thresholds.rs
```diff
@@ -0,0 +1,47 @@
+use std::{fs, path::Path};
+
+use crate::err_str;
+
+const START_THRESHOLD: &str = "/sys/class/power_supply/BAT0/charge_control_start_threshold";
+const END_THRESHOLD: &str = "/sys/class/power_supply/BAT0/charge_control_end_threshold";
+const UNSUPPORTED_ERROR: &str = "Not running System76 firmware with charge threshold suppport";
+const OUT_OF_RANGE_ERROR: &str = "Charge threshold out of range: should be 0-100";
+const ORDER_ERROR: &str = "Charge end threshold can't be less than start";
```
leviport, line 9: This works well,  but it looks like a `\` is getting inserted by the apostrophe:
```
system76@pop-os:~$ system76-power charge-thresholds 90 80
daemon returned an error message: "Charge end threshold can\'t be less than start"
```
We could change it to "cannot" if that's easier than figuring out where it's coming from.

leviport (changes requested):
Other than the above nitpick, this is working like a charm on oryp6! Once that's fixed, I'm ready to approve.

src/charge_thresholds.rs
ids1024: It seems the dbus-rs `Error` type is formatted in a rather strange way. I might send a PR upstream to improve that, but for now I've pushed a fix to this branch.

leviport:
The error message was fixed, but I just noticed that thresholds get reset to 0 100 if I power the machine off and unplug the charger. To me that sounds like it could be more of an issue with the oryp6 firmware changes we made though. Does that sound accurate?

ids1024:
As I understand, the EC firmware is supposed to persist the setting, but I guess it doesn't currently. In which case, there's nothing to change for that here.

@crawfxrd Does that sound accurate?

crawfxrd:
Correct. Only the keyboard config is persistent right now. The battery thresholds will be reset on EC reset.

leviport (dismissed):
Since it's an EC limitation, I think this is good to go. I'll open an issue in the EC repo to keep it on the radar.

ids1024:
Testing this in Gnome Control Center, I noticed it was silently failing when Polkit doesn't grant permission to set the threshold (for instance, if the dialog is dismissed). I've changed it to produce an error in that case.

(Probably different errors should be using different DBus error types, but to handle that properly, the way errors are done for dbus methods should probably all change.)

jacobgkau:
If I try to set the start and end thresholds to the same value, only the start is set:

```
system76@pop-os:~$ system76-power charge-thresholds
Start: 0
End: 90
system76@pop-os:~$ system76-power charge-thresholds 50 50
system76@pop-os:~$ system76-power charge-thresholds
Start: 50
End: 90
```

Similar to the case Levi pointed out before (but not quite the same), should this show an error message since it's not a valid configuration? (The error message could be changed from "Charge end threshold can't be less than start" to "Charge end threshold must be greater than start" if we want to use the same message for both cases.)

bflanagin (changes requested):
See Jacob's comment above.

Also, I noticed that we have 3 presets listed in the Gnome Control Center interface could those be added to the CLI similar to how we can set profiles?

ids1024:
Ah, I guess the EC implementation doesn't allow them to have the same value.

Updated with the error "Charge end threshold must be strictly greater than start" , and squashed down to one commit.

ids1024:
> Also, I noticed that we have 3 presets listed in the Gnome Control Center interface could those be added to the CLI similar to how we can set profiles?

I suppose it could be useful to support that. As implemented currently, `system76-power` provides the CLI and DBus API to set any valid values as thresholds, then gnome-control-center has the logic to set those based on profiles with names and descriptions.

bflanagin (approved):
Everything seems to be working as expected

Tests:
Start > End
Start == End
dbus: OK 
Changes persist through reboot: True (lemp9)

Unless you want to add my suggestion from yesterday (completely up to you). I think its good.

ids1024:
Supporting profiles names here is somewhat non-trivial. I don't think it's good to duplicate the profile names/values/descriptions in two places. So if we want to support that, I'd move that configuration here and add a DBus method to retrieve the list of profiles.

ids1024:
Ok, I've now added a DBus method to get a list of profiles, rewritten the gnome-control-center code to use this instead of hard-coding profiles, and added support for profile names in the CLI.

`system76-power charge-thresholds --profile full_charge` (or `balanced`, or `max_lifespan`).

jacobgkau:
If I just run `system76-power charge-thresholds`, I'm getting a very long string:

Screenshot from 2021-01-25 15-41-47

Is that supposed to be shown like this, or should it be formatted or hidden? (The bottom three lines look normal.)

ids1024:
Oops, seems I failed to remove a couple debug prints from when I was trying to beat dbus-rs into compliance. Amended.

jacobgkau:
The usage information is inconsistent between an invalid value and `--help`. For `--help`, I get:

```
USAGE:
    system76-power charge-thresholds [<start> <end>]
```

For an invalid value (such as `--profile` with no profile name, or a single number), I get:

```
USAGE:
    system76-power charge-thresholds <thresholds|--profile <profile>>
```

It would be more clear to specify `<start> <end>` in the latter case too, since that's what shows if someone tries to use a single number (i.e. a start but no end.) So something like this for both cases would cover all of the bases:

```
USAGE:
    system76-power charge-thresholds [<start> <end> | --profile <profile>]
```

Also, I don't see a way to list the available charging profiles. I'm able to list available power profiles with `system76-power profile --help` and available graphics profiles with `system76-power graphics --help` (the former only lists the available profiles, the latter provides a brief description of each available profile, so not entirely consistent but at least something for each.)

ids1024:
> The usage information is inconsistent between an invalid value and --help.

Interesting. `clap`, the library used here for argument parsing, doesn't seem to handle mutually exclusive arguments of this sort particularly well.

There probably should be a way to get a list of profiles through the CLI.

ids1024:
I've updated so that the help text lists the possible profile IDs, and for good measure also added a `--list-profiles` that lists those along with the title, description, and start/end thresholds.

I've also overridden the usage string for the subcommand to `system76-power charge-thresholds [<start> <end> | --profile <profile>]`, since the auto-generated one from clap seems to have some inconsistency as stated above.

jacobgkau:
The help text is looking good. I'm seeing strange behavior when attempting to set thresholds (only one threshold is being set in some situations.)

Example:
- Set thresholds to 40 and 60
- Set profile to `full_charge` (which should be 96 and 100)
- The result is thresholds of 40 and 100
- Setting the profile to `full_charge` a second time sets the start threshold to 96

Screenshot from 2021-01-26 09-48-01

Another example:
- Set profile to `max_lifespan` (50 and 60)
- Set thresholds to 70 and 80
- The result is thresholds of 50 and 80
- Setting to 70 and 80 a second time sets the start threshold to 70

Screenshot from 2021-01-26 09-45-51

This seems to happen when the new start threshold is greater than the old end threshold. (If the daemon is trying to set the thresholds in the same order every time, it would fail in those situations, because the firmware won't accept the higher start threshold until after the higher end threshold is set.)

ids1024:
Good catch.

> This seems to happen when the new start threshold is greater than the old end threshold. (If the daemon is trying to set the thresholds in the same order every time, it would fail in those situations, because the firmware won't accept the higher start threshold until after the higher end threshold is set.)

Yep. I wonder if the way the firmware is behaving is ideal here (I would compare how Lenovo etc. implement it if I had the hardware or documentation to do so).

Updated with a method that should fix this:
* Set end threshold to 100
* Set start threshold to requested value
* Set end threshold to requested value

jacobgkau (approved):
This is now working for all profiles and custom thresholds that I can throw at it. I noticed that attempting to pass more than two numbers gives a different error message based on whether I provide an even or odd number of parameters:

Screenshot from 2021-01-26 10-33-05

But the thresholds are not modified in either case, so I don't see that causing any issues. Thresholds are only modified when valid parameters are provided. The three new dbus methods are also working with gdbus, and regression testing passed.

ids1024:
That assertion should be impossible to reach since the argument parsing should have caught it before then. I guess adding it was a good idea.

Seems to be a bug in clap: https://github.com/clap-rs/clap/issues/2229

Amended to pass `.max_values(2)` which seems to make it behave more correctly.

jacobgkau (dismissed):
The error message still alternates, but the assertion failure is gone.

Screenshot from 2021-01-26 11-26-42

Still working as expected after the last change.

ids1024:
Yeah, still not quite right, but that should be good enough. Hopefully a future clap release fixes that bug.

jackpot51 (dismissed):

jackpot51:
I think this should land after #209 and it should also bump the version

jackpot51 (approved):

PR #198, opened 2020-10-28
fix: Set NVreg_PreserveVideoMemoryAllocations=1 on Nvidia and Hybrid

ids1024:
Hopefully this should help with the suspend/resume issues with WebRender. Requires testing to make sure it fixes the issue and doesn't cause others.

Related: https://github.com/pop-os/default-settings/pull/104 and https://github.com/pop-os/default-settings/pull/105.


jackpot51:
These settings, I believe, should go in https://github.com/pop-os/nvidia-graphics-drivers/blob/master/debian/templates/nvidia-graphics-drivers.conf.in and https://github.com/pop-os/nvidia-graphics-drivers/blob/master/debian/nvidia-graphics-drivers.conf

ids1024:
Well, that's easy enough, anyway.

Closing in favor of https://github.com/pop-os/nvidia-graphics-drivers/pull/70.

PR #204, opened 2020-12-17
Add GetDefaultGraphics DBus method

crawfxrd:
Add a new DBus method to get the graphics mode a product should use by default.
    
Only System76 models will default to "hybrid"; Other vendors will default to "nvidia". Some models, such as those without NVIDIA RTD3 power management, should default to "integrated" instead.

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -51,6 +51,8 @@ pub enum GraphicsDeviceError {
     Command { cmd: &'static str, why: io::Error },
     #[error(display = "{} in use by {}", func, driver)]
     DeviceInUse { func: String, driver: String },
+    #[error(display = "failed to read DMI info: {}", _0)]
+    Dmi(io::Error),
     #[error(display = "failed to open system76-power modprobe file: {}", _0)]
     ModprobeFileOpen(io::Error),
     #[error(display = "failed to write to system76-power modprobe file: {}", _0)]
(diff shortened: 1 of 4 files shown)


src/graphics.rs
```diff
@@ -219,6 +221,31 @@ impl Graphics {
...
+        let product = fs::read_to_string("/sys/class/dmi/id/product_version")
+            .map_err(GraphicsDeviceError::Dmi)
+            .map(|s| s.trim().to_string())?;
+
+        // Only default to hybrid on System76 models
+        let vendor = fs::read_to_string("/sys/class/dmi/id/sys_vendor")
+            .map_err(GraphicsDeviceError::Dmi)
+            .map(|s| s.trim().to_string())?;
+
+        if DEFAULT_INTEGRATED.contains(&product.as_str()) || vendor != "System76" {
```
jackpot51, line 242: If vendor != System76, default mode should be nvidia. These strings should also match what get_vendor returns

src/graphics.rs
```diff
@@ -219,6 +221,33 @@ impl Graphics {
...
+            .map(|s| s.trim().to_string())?;
+
+        // Only default to hybrid on System76 models
+        let vendor = fs::read_to_string("/sys/class/dmi/id/sys_vendor")
+            .map_err(GraphicsDeviceError::Dmi)
+            .map(|s| s.trim().to_string())?;
+
+        if DEFAULT_INTEGRATED.contains(&product.as_str()) {
+            Ok("integrated".to_string())
+        } else if vendor != "System76" {
```
jackpot51, line 244: I'd put this test first, in the unlikely event some other vendor uses the same product version

jackpot51 (commented):
This looks good now. I have only one concern, will this work on the older oryp1, oryp2, and oryp3?

There were also even older products that had both Intel and NVIDIA graphics. I don't have an exhaustive list

crawfxrd:
It should preserve the current behavior of distinst, which only has orpy4.

The logic could be improved by checking `supported-gpus.json`:

- If it exists and the card supports `runtimepm`, use hybrid
- If it exists and the card doesn't support `runtimepm` (or is not listed), use integrated
- If it doesn't exists, use integrated

jackpot51 (approved):

jacobgkau (approved):
dbus response is working with `dbus-send`, I can see that it returns `integrated` on galp5 and `hybrid` on gaze15. On lemp10, returns an error `does not have switchable graphics`.

PR #206, opened 2020-12-24
Determine default graphics by runtimepm support

crawfxrd:
Use the `supported-gpus.json` file provided by the NVIDIA driver to check if the GPU supports the "runtimepm" feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -226,6 +226,11 @@ dependencies = [
 ]
 
 [[package]]
+name = "itoa"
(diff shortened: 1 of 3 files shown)


jacobgkau (approved):
Returning `hybrid` on addw2, `integrated` on galp5 (blacklisted), and `integrated` on oryp4 (not blacklisted). The new `supported-gpus.json` check is working as intended.

ids1024 (approved):

Issue #210, opened 2021-01-25
galp5 performance mode should be tuned


curiousercreative:
**Distribution (run `cat /etc/os-release`):**
 ```
NAME="Pop!_OS"
VERSION="20.10"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 20.10"
VERSION_ID="20.10"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=groovy
UBUNTU_CODENAME=groovy
LOGO=distributor-logo-pop-os
```
 
 **Issue/Bug Description:**
With performance power profile selected on AC power for i7-1165G7:
- TDP boosts to 44-45W for up to 28s (observed with [pcm](https://github.com/opcm/pcm))
- Thermal throttling at 88C
- TDP normalizes to 28W
- all-core frequency 3.6-3.8GHz w/ adequate cooling

Writing MSRs (93C, pl1 45W 99999) results in:
- TDP boosts to 44-45W indefinitely (as thermals allow)
- Thermal throttling at 93C (exceeding 95C seems to trigger some cooldown)
- all-core frequency 4.0-4.1GHz w/ adequate cooling (meets spec)

**Other Notes:**
- [GB5 Multi-core score of 5704 with tuning](https://browser.geekbench.com/v5/cpu/6106942)
- [GB5 Multi-core score of 5583 baseline](https://browser.geekbench.com/v5/cpu/5974525)
- Similar issue that was resolved: https://github.com/pop-os/system76-power/issues/166
- [Battery power can achieve similar results with power limit increased in EC](https://github.com/curiousercreative/ec/commit/cbed54e4c515f1afeffa19bbe0a0000d09e02cfa#diff-198f0231dadddcc1692e1557f98782a282a0e540eb1664fd95402cb0574caab8R37)
- [performance tuning most useful with improved thermal management](https://github.com/system76/ec/pull/139)
- haven't discovered exactly how, but either connecting or disconnecting from AC power triggers TDP limit of 35-36W that can't be overcome with MSR until next cold boot, likely firmware related.

jackpot51:
I believe a TDP of 40W is acceptable for the galp5 on performance mode. The limit you see when on DC is to prevent overdrawing the battery. It is critical that this limit remains in place.

[...]

PR #212, opened 2021-01-26
galp5 profiles and no longer require all profile attributes to be set

curiousercreative:
Fixes https://github.com/pop-os/system76-power/issues/210

Observations from running these profiles on an i7 galp5 w/o dGPU:
- performance mode will allow for indefinite turbo boost up to all-core 4.1GHz, but thermally limited
- allowing temps to exceed 95C seems to trigger some cooldown period
- setting pl2 seems to disable or shorten the pl2 boost duration such that the performance profile is useless
- it seems sufficient to not define pl1 or pl2 at all and just let thermal limiting handle everything

### DIFF
--- a/src/daemon/profiles.rs
+++ b/src/daemon/profiles.rs
@@ -196,9 +196,9 @@ fn set_disk_power(apm_level: u8, autosuspend_delay: i32) -> Result<(), DiskPower
 }
 
 pub struct ModelProfile {
-    pl1: u8,
-    pl2: u8,
-    tcc_offset: u8,
+    pl1: Option<u8>,
+    pl2: Option<u8>,
+    tcc_offset: Option<u8>,
 }
 
 impl ModelProfile {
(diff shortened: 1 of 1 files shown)


src/daemon/profiles.rs
```diff
@@ -211,19 +211,23 @@ impl ModelProfile {
         //TODO: check status, allow thermald to be missing
 
         // Set PL1
-        fs::write(
-            "/sys/class/powercap/intel-rapl:0/constraint_0_power_limit_uw",
-            format!("{}", (self.pl1 as u64) * 1_000_000)
-        ).map_err(ModelError::Pl1)?;
+        if self.pl1.is_some() {
```
jackpot51, line 214: This could be changed to:

if let Some(pl1) = self.pl1 {
   ... (pl1 as u64)...
}

src/daemon/profiles.rs
```diff
@@ -211,19 +211,23 @@ impl ModelProfile {
...
+                format!("{}", (self.pl1.unwrap() as u64) * 1_000_000)
+            ).map_err(ModelError::Pl1)?;
+        }
 
         // Set PL2
-        fs::write(
-            "/sys/class/powercap/intel-rapl:0/constraint_1_power_limit_uw",
-            format!("{}", (self.pl2 as u64) * 1_000_000)
-        ).map_err(ModelError::Pl2)?;
+        if self.pl2.is_some() {
```
jackpot51, line 222: And a similar change here

src/daemon/profiles.rs
```diff
@@ -211,19 +211,23 @@ impl ModelProfile {
...
+            fs::write(
+                "/sys/class/powercap/intel-rapl:0/constraint_1_power_limit_uw",
+                format!("{}", (self.pl2.unwrap() as u64) * 1_000_000)
+            ).map_err(ModelError::Pl2)?;
+        }
 
         // Set TCC
-        {
+        if self.tcc_offset.is_some() {
```
jackpot51, line 230: And a similar change here

curiousercreative:
@jackpot51 thanks for the feedback, updated with your suggestions and tested on my galp5.

jackpot51 (approved):

jacobgkau (approved):
Before changes, `constraint_0_power_limit_uw` was always `28000000` and `constraint_1_power_limit` was always `51000000` on galp5. After changes, `constraint_1` still doesn't change, but `constraint_0` gets set to `12000000` in `battery`, `28000000` in `balanced`, and `40000000` in `performance`.

While stressing the galp5, I observed an energy rate of 66W (which then dropped to 52W after a few minutes) in `performance`. Switching to `battery` caused a dramatic decrease in energy rate to 19W, increasing the estimated remaining battery time from 49 minutes to 2.2 hours. Switching back to performance brought the energy rate back up. This seems to indicate that the options are working as intended. (Before the change, the rate was around 39W in `performance` and 28W in `battery`, not as effective.)

On lemp9, both power limits are still set when switching profiles; on gaze15, both limits are still left unchanged. General regression testing passed.

ZeddieXX:
I just want to clarify, the truly broken experience seem to be the Battery Mode.  When set to Battery Mode and CPU is at high utilization, it starts at 2.4 GHz, but then slowly ramps down to 218 MHz!  (2.2x multiplier).

I don't think that was intended.  If the CPU can clock down to 2.2x (I didn't think that was possible, per Intel's spec minimum is 4x), it should only do so at or near idle, not when there is load.

jackpot51:
Thanks for that, it should be easy to adjust. I will make a PR soon.

curiousercreative:
@jackpot51 for my own curiosity, why would it be a bad idea to maintain the same pl1 (40) for all profiles and just vary the thermal limit?

jackpot51:
Mainly to prevent runaway thermal issues that would cause "sputtering", similar to what @ZeddieXX is experiencing but worse.

It would run 40W until it hits the thermal limit, which would happen very quickly. Then it would stall and fall down to very low wattage. Then it cools and hits 40W again, the cycle repeating.

PR #213, opened 2021-01-27
Report default graphics as integrated if checking RTD3 support fails

crawfxrd:
If the NVIDIA device is not on the bus or the drivers are not loaded, assume run-time power management is not supported.

Tested with pop-os/gnome-initial-setup#68.

Fixes: 8959756fee2 ("Determine default graphics by runtimepm support")

### DIFF
--- a/data/system76-power.conf
+++ b/data/system76-power.conf
@@ -15,4 +15,8 @@
         <allow send_destination="com.system76.PowerDaemon"/>
         <allow receive_sender="com.system76.PowerDaemon"/>
     </policy>
+    <policy user="gnome-initial-setup">
+        <allow send_destination="com.system76.PowerDaemon"/>
+        <allow receive_sender="com.system76.PowerDaemon"/>
+    </policy>
 </busconfig>
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -312,6 +312,10 @@ impl Graphics {
             .map(|s| s.trim().to_string())?;
         let blacklisted = DEFAULT_INTEGRATED.contains(&product.as_str());
 
+        // If the NVIDIA device is not on the bus or the drivers are not
+        // loaded, then assume runtimepm is not supported.
+        let runtimepm = self.gpu_supports_runtimepm().unwrap_or_default();
+
         // Only default to hybrid on System76 models
         let vendor = fs::read_to_string("/sys/class/dmi/id/sys_vendor")
             .map_err(GraphicsDeviceError::SysFs)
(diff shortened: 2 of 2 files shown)


jackpot51 (dismissed):

jacobgkau:
This gets the GetDefaultGraphics method working on galp5, where it now returns `integrated` whether I'm booted into Integrated, Hybrid, or NVIDIA mode. However, on addw2, I get `integrated` if I'm booted into Integrated mode (but I get the correct response of `hybrid` if I'm booted into Hybrid or NVIDIA modes.)

This seems like it shouldn't really be an issue, as the ISO boots into NVIDIA mode (where the response should be correct) and from there it installs into Hybrid mode on systems that default to Hybrid (so the response should also be correct on first boot.) However, the relevant page of GNOME Initial Setup does show up when a new user logs in for the first time, and if that happens in Integrated mode, the new user will see a panel stating the machine defaults to Integrated mode.

Just to confirm, is this corner case acceptable, or are there any other problems it could cause? It seems like GIS is still doing its job explaining how to switch graphics modes from the current mode, but at that point, it seems like GIS could just be using the current mode instead of the default mode to determine what to show on this page.

crawfxrd:
It won't cause any problems since distinst and g-i-s are the only things depending on it. The only thing that could happen is, as you point out, if you explicitly change to integrated and then add a new user, it reports integrated instead of hybrid. (And only on System76 models, as other vendor models will continue reporting NVIDIA as the default mode.)

To improve this would require determining the version of the NVIDIA drivers installed without them being loaded. I wasn't sure of a simple way to do this though, so it didn't seem worth trying to address yet.

jacobgkau (approved):
As discussed, the last commit was necessary to get GNOME Initial Setup working in the first boot environment (where GIS is being run as the `gnome-initial-setup` user, which is not part of the `sudo` or `adm` groups.)

Testing together with https://github.com/pop-os/gnome-initial-setup/pull/68:

galp5 NVIDIA:
- In live environment (booted in NVIDIA mode), dbus GetDefaultGraphics returns `integrated`.
- On first boot (Integrated mode), GNOME Initial Setup displays the correct "defaults to Integrated graphics" text.

gaze15:
- In live environment (booted in NVIIDA mode), dbus GetDefaultGraphics returns `hybrid`.
- On first boot (Hybrid mode), GNOME Initial Setup displays the correct "defaults to Hybrid graphics" text.

jackpot51 (approved):

PR #217, opened 2021-02-02
1.1.14 - Add thelio-major-b3

jackpot51:
Rebase and tag when approved. This sets it up to use the hedt fan curve like thelio-major-b2


leviport (approved):
I haven't tested this fan curve with a built-out chassis, but the major-b2 curves seem like they'll be just fine.

PR #224, opened 2021-03-22
chore(debian): Update use of debhelper

ids1024:
Update to compat level 11 (newest supported by bionic). `--with=systemd` is now implied.

Also fix uses of `shlibs`, which adds missing library dependencies.


jackpot51 (approved):

jacobgkau (approved):
Regression testing passed.

PR #230, opened 2021-04-04
use declarative Clap syntax

### LINKED ISSUE #226: Declarative command line arguments
would you accept a pull request for converting the command-line argument parsing to the declarative, derive-based style in StructOpt or Clap 3.0-beta?

danieleades:
closes #226


src/cli/charge_thresholds.rs
```diff
@@ -0,0 +1,78 @@
+use clap::Clap;
+use system76_power::{charge_thresholds::get_charge_profiles, client::PowerClient};
```
danieleades, line 2: should i be using `system76_power::charge_thresholds::get_charge_profiles` or `system76_power::client::PowerClient::get_charge_profiles`??

what's their relationship?

Cargo.toml
```diff
@@ -9,7 +9,7 @@ dbus = "0.9.0"
 dbus-crossroads = "0.3.0"
 dbus-tokio = "0.6.0"
 libc = "0.2.76"
-clap = "2.33.3"
+clap = "=3.0.0-beta.2"
```
crawfxrd, line 12: Might be better to pin a recent git commit. The crate is 7 months old.

danieleades: with the exception of the `arg_enum` annotation, i'm not using any functionality that hasn't existed for several years under `structopt`. I don't subscribe to the notion that a newer commit is *necessarily* more stable than the most recent published release, but i'm happy to be persuaded otherwise

crawfxrd: Basically, just looking to minimize the overhead of updating to v3 if it ever releases. It looks like there won't be another tag and the maintainer recommends pinning instead for using a newer version (clap-rs/clap#2379).

danieleades: that was prescient of you-

```
error: failed to select a version for `proc-macro2`.
    ... required by package `clap_derive v3.0.0-beta.2 (https://github.com/clap-rs/clap.git?rev=d253c34676acf73c7ad04bcec56ff573e925681d#d253c346)`
    ... which is depended on by `clap v3.0.0-beta.2 (https://github.com/clap-rs/clap.git?rev=d253c34676acf73c7ad04bcec56ff573e925681d#d253c346)`
    ... which is depended on by `system76-power v1.1.16 (/home/dan/Code/rust/system76-power)`
versions that meet the requirements `^1.0.26` are: 1.0.26

all possible versions conflict with previously selected packages.

  previously selected package `proc-macro2 v1.0.24`
    ... which is depended on by `err-derive v0.2.4`
    ... which is depended on by `intel-pstate v0.2.1`
    ... which is depended on by `system76-power v1.1.16 (/home/dan/Code/rust/system76-power)`

failed to select a version for `proc-macro2` which could resolve this conflict
```

probably upgrading `intel-pstate` to use `thiserror` instead of `err-derive` would do the trick. but in order to do that, first i need to shave the yak...

danieleades: see https://github.com/pop-os/intel-pstate/pull/2

danieleades: - `err-derive` depends on proc-macro2 `^1.0.1`
- `thiserror` depends on proc-macro2 `^1.0`
- clap depends on proc-macro2 `^1.0.26`

i'm not entirely sure why proc-macro2 version 1.0.26 can't satisfy everyone...
i'm also unclear on whether my pull request on intel-pstate is going to help the situation

crawfxrd: I think err-derive is deprecated, so that change is still good regardless.

The version of proc-macro2 in the lock file is 1.0.19, so perhaps a `cargo update -p proc-macro2` would resolve it.

danieleades: the remaining build errors are due to `sysfs-class` making a breaking change in this commit - https://github.com/pop-os/sysfs-class/commit/ab63e7f638aadfaf896a02e53cf330343d331337

`system76-power` should pin the sysfs-class version it's using

ids1024:
Clap 3.0 has finally been released.

danieleades:
this is out of date

PR #233, opened 2021-04-06
run tests in CI

danieleades:
the test coverage is obviously pretty minimal, but what's there might as well be tested.


danieleades:
rebased

danieleades:
rebased on master.

any interest in merging this? @crawfxrd

danieleades:
@crawfxrd i'm not sure if you have an internal process for skipping QA on some pull requests, but this one (and also #238) don't contain any code changes at all, so shouldn't need to be tested beyond observing that the corresponding [github action](https://github.com/pop-os/system76-power/pull/233/checks?check_run_id=3125629772) runs (which it does)

jacobgkau:
It looks like the new check is essentially just running `cargo test --all --all-features` and reporting the warnings/errors? Perhaps the check should be called `Cargo Test` instead of just `Test`?

danieleades:
> It looks like the new check is essentially just running `cargo test --all --all-features` and reporting the warnings/errors? Perhaps the check should be called `Cargo Test` instead of just `Test`?

I can change it to `Cargo Test` if you prefer. Whatever makes the most sense for you guys.

`Cargo` is really just the entry point though. the rustfmt and clippy jobs aren't prefixed with `cargo`, even though they are invoked using Cargo. And there are a lot of other jobs you could (in theory) add- security advisory audit, transitive licence checking, code coverage (#238), all of which use cargo as the entry point. I guess in the case of the tests you're hoping to distinguish the manual tests from the unit, integration, documentation, and bench tests that `cargo test` will run?

danieleades:
i'm just removed the `--all` flag. It won't hurt, but it also won't do anything in this project since it doesn't have a virtual manifest. It's been deprecated in favour of `--workspace` anyway

jacobgkau:
> I can change it to `Cargo Test` if you prefer. Whatever makes the most sense for you guys.
> 
> `Cargo` is really just the entry point though. the rustfmt and clippy jobs aren't prefixed with `cargo`, even though they are invoked using Cargo.

I just thought that `Test` was an incredibly vague title that does not describe anything about what test is being run. I had to read `CI.yml` and find `uses: actions-rs/cargo@v1` to understand what command it's running. However, if this is following the naming convention that the other tests are already using, that's fine.

jacobgkau (approved):
No QA concerns with this, engineering can decide if it's necessary.

jackpot51 (approved):

PR #234, opened 2021-04-06
test using 'cargo' action instead of 'clippy-check'

danieleades:
i've used the 'clippy-check' action instead of the 'cargo' action in the github workflow because the 'clippy-check' action is supposed to be able to comment inline on pull requests.

that doesn't work on pull requests from forks, which is a major limitation. however this *did* work in the 'test' action in #233, which is surprising.

i've created this draft pull request to compare the two actions


danieleades:
this doesn't help. closing

PR #246, opened 2021-04-30
Document that a reboot is required after switching

crawfxrd:
Switching graphics modes requires a reboot for changes to take effect as it modifies module loading rules that happen during early boot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 
 ## Graphics Modes
 
+A reboot is **required** for changes to take effect after switching modes.
+
 ### Integrated
 
 The integrated graphics controller on the Intel or AMD CPU is used exclusively.
--- a/src/client.rs
+++ b/src/client.rs
@@ -84,7 +84,11 @@ impl Power for PowerClient {
 
     fn set_graphics(&mut self, vendor: &str) -> Result<(), String> {
         println!("setting graphics to {}", vendor);
-        self.call_method::<&str>("SetGraphics", Some(vendor)).map(|_| ())
+        let r = self.call_method::<&str>("SetGraphics", Some(vendor)).map(|_| ());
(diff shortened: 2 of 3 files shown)


bflanagin (approved):
No regressions detected.

jacobgkau:
The help text looks fine:

Screenshot from 2021-04-30 10-52-45

Just a thought, would it make sense to instead/also show a message about rebooting after a graphics-switching command has been run, like `___ graphics will be enabled on the next reboot.`? This would match the GUI behavior closer.

crawfxrd:
Switching now tells the user to reboot if it is successful.

MilesBHuff:
@crawfxrd Any chance you could add a message to the getter as well (if a reboot is needed)?

MilesBHuff:
Actually, something else just occurred to me:  hybrid -> compute is possible without a reboot.  The warning messages here don't account for that.

crawfxrd:
The daemon doesn't track the state (it's calculated on every query), so the getter can't report if a reboot is required.

Hybrid <-> Compute still requires restarting the X server. It can be done nicely on GNOME 3.38 since GDM isn't kept alive on vt1 [\[1\]](https://github.com/pop-os/system76-power/issues/57#issuecomment-700027201). But we still tell users to reboot.

MilesBHuff:
Interesting!  Okay.  In my experience, switching from hybrid to integrated returns "compute" in the getter.
```
swenthorian@oryp7:~/
% sudo system76-power graphics
hybrid

swenthorian@oryp7:~/
% sudo system76-power graphics integrated
setting graphics to integrated

swenthorian@oryp7:~/
% sudo system76-power graphics           
compute
```

jackpot51 (approved):

PR #257, opened 2021-07-12
Update deps, fix building

crawfxrd:
Update dependencies and fix building with latest sysfs-class.


jackpot51 (approved):

jacobgkau (approved):
Regression testing passed, graphics, power, and charging profiles still work.

PR #258, opened 2021-07-12
Add gaze16

crawfxrd:


jackpot51 (approved):
I can confirm the 3060 pins are correct

jacobgkau:
Hotplug detection works for Mini DisplayPort on gaze16-3050 (and HDMI output works in Integrated mode.) Graphics switching also works.

This is still marked as draft, is it ready for review or are further changes anticipated?

jacobgkau (approved):

PR #269, opened 2021-09-14
Add oryp8

crawfxrd:


jackpot51 (approved):

jacobgkau (approved):
Graphics switching between all modes works, and hotplugging displays works as expected in all modes (Integrated & Compute show prompt to switch to Hybrid, Hybrid and NVIDIA work without prompt.)

PR #271, opened 2021-09-17
chore(rust): Update MSRV to 1.51.0

### LINKED ISSUE #270: Compile issues related to clap
 obarun
 
 
system76-power v1.1.18

**Issue/Bug Description:**
The compile fails because of this issue https://github.com/clap-rs/clap/issues/2691 and my knowledge of rust is not high enough to see what needs to be done here :( seems like the rust version pinned by the project needs to be bumped?

Compile the package on an arch based distro

mmstick:
Closes #270


jackpot51 (approved):

jacobgkau (approved):
Regression testing passed:
[system76-power-pr271-hirsute.txt](https://github.com/pop-os/system76-power/files/7187981/system76-power-pr271-hirsute.txt)
[system76-power-pr271-focal.txt](https://github.com/pop-os/system76-power/files/7188299/system76-power-pr271-focal.txt)

danieleades:
Just a nit- the `rust-toolchain` file doesn't define an MSRV, it defines a pinned compiler version for development, which is not quite the same thing.
Are you *also* using it as a pinned compiler version for reproducible builds or *only* as the MSRV? if the latter, your be better off adding an MSRV target to the CI job, deleting the `rust-toolchain` file, and developing using the stable or nightly toolchain.

For example- https://github.com/danieleades/monzo-lib/blob/main/.github/workflows/CI.yml

mmstick:
This is our minimum-supported Rust compiler version, based on the version of rustc currently available on all of the Ubuntu releases we support. It ensures that when we are working on a project, rustup automatically uses this version of Rust so we don't have to wait for sbuild to complain about a compiler error.

danieleades:
@mmstick thanks for the explanation, that makes sense.

mmstick:
Yeah, our needs are a bit different from what most would do, since we have an internal packaging server as our CI that uses Ubuntu chroots to build each package and then generates a debian repositories for every branch. Sometimes a crate gets an update that uses a newer version of Rust than is available in Ubuntu, and our builds break, unfortunately.

danieleades:
> Yeah, our needs are a bit different from what most would do, since we have an internal packaging server as our CI that uses Ubuntu chroots to build each package and then generates a debian repositories for every branch. Sometimes a crate gets an update that uses a newer version of Rust than is available in Ubuntu, and our builds break, unfortunately.

fair enough. that must make bumping dependencies a bit of a crapshoot? hopefully the [rust-version](https://doc.rust-lang.org/cargo/reference/unstable.html#rust-version) feature gains enough traction to be useful. Would be nice if the version resolution could account for that for you.

PR #273, opened 2021-09-21
lint and fix with nightly clippy

danieleades:
- fixes hard errors in stable clippy causing CI to fail on master
- fixes nightly clippy lints (`cargo +nightly clippy`)
- introduces clippy config file to specify MSRV
- suppres 'dead code' warning in example

specifying the MSRV prevents clippy from throwing false positives for syntax that's newer than is supported by this crate. I've just chosen the MSRV to match the `rust-toolchain` file.


jacobgkau (approved):
Regression testing passed:
[system76-power-pr273-hirsute.txt](https://github.com/pop-os/system76-power/files/7239378/system76-power-pr273-hirsute.txt)
[system76-power-pr273-focal.txt](https://github.com/pop-os/system76-power/files/7239379/system76-power-pr273-focal.txt)

PR #278, opened 2021-10-06
feat: Use Linux's ACPI Platform Profiles when available

mmstick:
As far as I can tell, only Lenovo laptops support this today.

### DIFF
--- a/src/acpi_platform.rs
+++ b/src/acpi_platform.rs
@@ -1,0 +1,34 @@
+// Copyright 2018-2021 System76 <info@system76.com>
+//
+// SPDX-License-Identifier: GPL-3.0-only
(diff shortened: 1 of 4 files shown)


jackpot51 (approved):

jacobgkau (approved):
We don't have any Lenovo machines in the lab to test that the new functionality will work. Regression testing has passed on our products:
[system76-power-pr278-hirsute.txt](https://github.com/pop-os/system76-power/files/7312685/system76-power-pr278-hirsute.txt)
[system76-power-pr278-focal.txt](https://github.com/pop-os/system76-power/files/7313028/system76-power-pr278-focal.txt)

PR #279, opened 2021-10-07
add dependabot config

danieleades:
no code changes in this one

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+- package-ecosystem: cargo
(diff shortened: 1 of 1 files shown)


crawfxrd:
I find Dependabot rather annoying.

danieleades:
> I find Dependabot rather annoying.

Do you now? It can be configured to only make pull requests for security patches. Would that be preferable?

I find dependabot is only annoying when it's doing something useful. If your dependencies are up to date, it's silent

jackpot51:
@danieleades we would be okay with security updates only

Issue #276, opened 2021-10-02
Does not work on non LTS kernel


CrafterSvK:
**Distribution (run `cat /etc/os-release`):**
```
NAME="Arch Linux"
PRETTY_NAME="Arch Linux"
ID=arch
BUILD_ID=rolling
ANSI_COLOR="38;2;23;147;209"
HOME_URL="https://archlinux.org/"
DOCUMENTATION_URL="https://wiki.archlinux.org/"
SUPPORT_URL="https://bbs.archlinux.org/"
BUG_REPORT_URL="https://bugs.archlinux.org/"
LOGO=archlinux
```
 
 
**Related Application and/or Package Version (run `apt policy $PACKAGE NAME`):**
`aur/system76-power 1.1.18-1`

**Issue/Bug Description:**
Does not work on `testing` and `core` linux kernel on Arch. For a long time.

**Steps to reproduce (if you know):**
`# system76-power graphics power on`
`# system76-power graphics power`
`off (discrete)`
`# system76-power graphics nvidia`
Does not work after reboot (not even in lts kernel)

**Expected behavior:**
Turning graphics power on, actually turning it on, not just saying it did.

**Other Notes:**
oryp6 RTX 2060
Turning power manually on, works on lts kernel for a year.

jacobgkau:
This seems like a duplicate of https://github.com/pop-os/system76-power/issues/268, https://github.com/pop-os/system76-power/issues/267, and possibly also https://github.com/pop-os/system76-power/issues/265.

PR #281, opened 2021-10-12
fix: checks for missing entries, use sane defaults

zsxawerdu:
Resolves [https://github.com/pop-os/system76-power/issues/276](#276 )

### DIFF
--- a/src/hid_backlight.rs
+++ b/src/hid_backlight.rs
@@ -76,17 +76,43 @@ pub fn daemon() {
     let color_file = dir.join("color");
 
     let mut inotify = Inotify::init().unwrap();
-    inotify.add_watch(&brightness_file, WatchMask::MODIFY).unwrap();
-    inotify.add_watch(&brightness_hw_changed_file, WatchMask::MODIFY).unwrap();
-    inotify.add_watch(&color_file, WatchMask::MODIFY).unwrap();
+    let brightness_file = match brightness_file.exists() {
+        true => {
+            inotify.add_watch(&brightness_file, WatchMask::MODIFY).unwrap();
+            Some(brightness_file)
+        },
+        false => None,
+    };
+    let brightness_hw_changed_file = match brightness_hw_changed_file.exists() {
+        true => {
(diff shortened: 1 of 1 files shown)


src/hid_backlight.rs
```diff
@@ -76,17 +76,43 @@ pub fn daemon() {
...
-    inotify.add_watch(&brightness_file, WatchMask::MODIFY).unwrap();
-    inotify.add_watch(&brightness_hw_changed_file, WatchMask::MODIFY).unwrap();
-    inotify.add_watch(&color_file, WatchMask::MODIFY).unwrap();
+    let brightness_file = match brightness_file.exists() {
+        true => {
+            inotify.add_watch(&brightness_file, WatchMask::MODIFY).unwrap();
+            Some(brightness_file)
+        },
+        false => None,
+    };
```
mmstick, line 85: Matching a boolean is considered a bad pattern by cargo-clippy. `if condition { Some(thing) } else { None }` would be better.

src/hid_backlight.rs
```diff
@@ -76,17 +76,43 @@ pub fn daemon() {
...
     let mut buffer = [0; 1024];
     loop {
-        let brightness_string = fs::read_to_string(&brightness_file).unwrap();
-        let brightness = brightness_string.trim().parse::<u8>().unwrap();
-
-        let color_string = fs::read_to_string(&color_file).unwrap();
-        let color = u32::from_str_radix(color_string.trim(), 16).unwrap();
+        let brightness = if let Some(ref brightness_file) = brightness_file {
+            let brightness_string = fs::read_to_string(&brightness_file).unwrap();
+            brightness_string.trim().parse::<u8>().unwrap()
```
mmstick, line 105: I think it'd be better if we handled these errors by logging them rather than unwrapping with a panic.

src/hid_backlight.rs
zsxawerdu: I have a WIP branch with these changes.  I am currently QAing this locally before next commit.  https://github.com/zsxawerdu/system76-power/blob/WIP/experimental/src/hid_backlight.rs#L78

Issue #267, opened 2021-09-13
Unable to start daemon (hid_backlight error) on Lemur Pro


lumnn:
**Distribution (run `cat /etc/os-release`):**
 
Arch Linux
 
**Related Application and/or Package Version (run `apt policy $PACKAGE NAME`):**

```
aur/system76-driver 20.04.41-2 [installed] (21, 0.01)
aur/system76-power 1.1.17-1 [installed] (9, 0.11)
aur/system76-firmware 1.0.29-1 [installed] (9, 0.01)
aur/system76-firmware-daemon 1.0.29-1 [installed] (9, 0.01)
aur/system76-dkms 1.0.13-1 [installed] (6, 0.06)
aur/gnome-shell-extension-system76-power-git 2.0.1.r1.gfc4aca1-1 [installed] (5, 0.00)
aur/system76-io-dkms 1.0.1-1 [installed] (2, 0.00)
```

**Issue/Bug Description:**

```
● system76-power.service - System76 Power Daemon
     Loaded: loaded (/usr/lib/systemd/system/system76-power.service; enabled; vendor preset: disabled)
     Active: active (running) since Mon 2021-09-13 14:07:29 CEST; 10min ago
   Main PID: 664 (system76-power)
      Tasks: 9 (limit: 18982)
     Memory: 3.5M
        CPU: 300ms
     CGroup: /system.slice/system76-power.service
             └─664 /usr/bin/system76-power daemon

Sep 13 14:07:29 lem system76-power[664]: [INFO] Setting automatic graphics power
Sep 13 14:07:29 lem system76-power[664]: [WARN] Failed to set automatic graphics power: does not have switchable graphics
Sep 13 14:07:29 lem system76-power[664]: [INFO] Initializing with the balanced profile
Sep 13 14:07:29 lem system76-power[664]: [INFO] Registering dbus name com.system76.PowerDaemon
Sep 13 14:07:29 lem system76-power[664]: [INFO] Adding dbus path /com/system76/PowerDaemon with interface com.system76.PowerDaemon
Sep 13 14:07:29 lem system76-power[664]: [ERROR] fan daemon: platform hwmon not found
Sep 13 14:07:29 lem system76-power[664]: [INFO] Handling dbus requests
[...]

PR #290, opened 2021-11-05
fix: Handle lack of color on some keyboard LEDs

jacobgkau:
Fixes https://github.com/pop-os/system76-power/issues/267 by taking the potential lack of keyboard backlight color file into account (for products such as lemp10, which have a single-color keyboard backlight.)

There might be a better way to do this; this seemed like the method that would require the least amount of changes to the code.

Not worrying about handling lack of brightness files right now because I'm not aware of any System76 products with keyboards that have LEDs that don't have controllable brightness. If a product doesn't have a keyboard backlight at all, then we already return before we get to this point.

### DIFF
--- a/src/hid_backlight.rs
+++ b/src/hid_backlight.rs
@@ -78,14 +78,18 @@ pub fn daemon() {
     let mut inotify = Inotify::init().unwrap();
     inotify.add_watch(&brightness_file, WatchMask::MODIFY).unwrap();
     inotify.add_watch(&brightness_hw_changed_file, WatchMask::MODIFY).unwrap();
-    inotify.add_watch(&color_file, WatchMask::MODIFY).unwrap();
+    if let Err(e) = inotify.add_watch(&color_file, WatchMask::MODIFY) {
+        log::warn!("hid_backlight: failed to watch keyboard color: {}", e);
+    }
 
     let mut buffer = [0; 1024];
     loop {
         let brightness_string = fs::read_to_string(&brightness_file).unwrap();
         let brightness = brightness_string.trim().parse::<u8>().unwrap();
 
-        let color_string = fs::read_to_string(&color_file).unwrap();
+        #[rustfmt::skip]
+        let color_string = fs::read_to_string(&color_file)
+            .unwrap_or_else(|_| String::from("FFFFFF")); // Fallback for non-colored keyboards
         let color = u32::from_str_radix(color_string.trim(), 16).unwrap();
 
         let mut devices = 0;


src/hid_backlight.rs
```diff
@@ -70,22 +70,26 @@ pub fn daemon() {
...
-    // TODO: check for existence of files
     let brightness_file = dir.join("brightness");
     let brightness_hw_changed_file = dir.join("brightness_hw_changed");
     let color_file = dir.join("color");
 
     let mut inotify = Inotify::init().unwrap();
     inotify.add_watch(&brightness_file, WatchMask::MODIFY).unwrap();
     inotify.add_watch(&brightness_hw_changed_file, WatchMask::MODIFY).unwrap();
-    inotify.add_watch(&color_file, WatchMask::MODIFY).unwrap();
+    if Path::new(&color_file).is_file() {
```
ids1024, line 80: This should work, but it would probably be better to test the return value of `add_watch` and `read_to_string` instead of wrapping lines using `.unwrap()` in an `if` block.

jacobgkau: Got it, I saw people recommending handling the error instead when I looked up how to do this. Will try to get it working that way.

src/hid_backlight.rs
```diff
@@ -70,22 +70,26 @@ pub fn daemon() {
...
     let mut buffer = [0; 1024];
     loop {
         let brightness_string = fs::read_to_string(&brightness_file).unwrap();
         let brightness = brightness_string.trim().parse::<u8>().unwrap();
 
-        let color_string = fs::read_to_string(&color_file).unwrap();
+        let mut color_string = String::from("FFFFFF"); // Fallback for non-colored keyboards
+        if Path::new(&color_file).is_file() {
+            color_string = fs::read_to_string(&color_file).unwrap();
+        }
```
crawfxrd, line 92: Can use `unwrap_or_else()` to return a default value. Keep the comment explaining the value.

```diff
- .unwrap();
+ .unwrap_or_else(|_| String::from("FFFFFF"));
```

jacobgkau:
Here's how the logs look now on lemp10:
Screenshot from 2021-11-05 15-13-30

src/hid_backlight.rs
jacobgkau: Better?

jacobgkau:
I tried to split line 92 to make it not as long, but the rustfmt check didn't like that.

src/hid_backlight.rs
```diff
@@ -78,14 +78,18 @@ pub fn daemon() {
...
+    }
 
     let mut buffer = [0; 1024];
     loop {
         let brightness_string = fs::read_to_string(&brightness_file).unwrap();
         let brightness = brightness_string.trim().parse::<u8>().unwrap();
 
-        let color_string = fs::read_to_string(&color_file).unwrap();
+        let color_string =
+            fs::read_to_string(&color_file).unwrap_or_else(|_| String::from("FFFFFF")); // Fallback for non-colored keyboards
```
crawfxrd, line 92: I don't know what rustfmt is doing. Wrapping on chains work on literally everything else. Just stick a `#[rustfmt::skip]` above the assignment and give it sane wrapping.

jacobgkau: Used the skip to give it better wrapping.

jackpot51 (approved):

13r0ck (approved):
I am seeing the same as @jacobgkau [here](https://github.com/pop-os/system76-power/pull/290#issuecomment-962224218). This fixes #267

src/hid_backlight.rs
danieleades: Could you not use the `// fallback for...` comment to split the lines, by placing it between the chained methods?

PR #292, opened 2021-11-18
graphics: Drop NVreg_DynamicPowerManagement

crawfxrd:
Since 465.19.01, `NVreg_DynamicPowerManagement` has a new value, 0x03, that is the default [\[1\]].

> For Ampere or later notebooks with supported configurations, this
> value translates to fine-grained power control. For pre-Ampere
> notebooks, this value disables runtime D3 power management features.

ubuntu-drivers-common's gpu-manager also sets this value, and will need to be updated to remove it as well (https://github.com/pop-os/ubuntu-drivers-common/pull/13).

Can be checked in the driver params:

    $ grep -w DynamicPowerManagement /proc/driver/nvidia/params
    DynamicPowerManagement: 3

Reported upstream: https://github.com/tseliot/ubuntu-drivers-common/issues/62

[\[1\]]: https://download.nvidia.com/XFree86/Linux-x86_64/465.19.01/README/dynamicpowermanagement.html#DriverSettingsfe624


jackpot51 (approved):

crawfxrd:
We probably shouldn't do this after all, since NVIDIA made it only apply to Ampere. RTD3 works on Turing cards, so I don't know why they excluded it...

13r0ck:
@crawfxrd do you still want us to QA this then?

crawfxrd:
No. If we have to keep it anyway, I'd rather do it here to avoid depending on ubuntu-drivers-common to do it.

crawfxrd:
Removing it in ubuntu-drivers-common should still done, since I want to remove the dependency on it eventually.

PR #293, opened 2021-11-18
graphics: Preserve video memory during sleep

crawfxrd:
The NVIDIA driver has two methods to save video memory during suspend:

- Save allocations in an unnamed temporary file
- S0ix-based power management

These are exclusive. Systems using S3 should only use the first, and
systems using S0ix should only use the second.

Requires: pop-os/nvidia-graphics-drivers#128
Ref: pop-os/nvidia-graphics-drivers@de515f69a481
Link: https://download.nvidia.com/XFree86/Linux-x86_64/470.86/README/powermanagement.html#PreserveAllVide719f0

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -48,6 +48,16 @@ alias nvidia-drm off
 alias nvidia-modeset off
 "#;
 
+// Systems using S0ix must enable S0ix-based power management.
+static SYSTEM_SLEEP_S0IX: &[u8] = br#"# Preserve video memory through suspend
+options nvidia NVreg_EnableS0ixPowerManagement=1
+"#;
+
+// Systems using S3 had suspend issues with WebRender.
+static SYSTEM_SLEEP_S3: &[u8] = br#"# Preserve video memory through suspend
+options nvidia NVreg_PreserveVideoMemoryAllocations=1
+"#;
+
 const PRIME_DISCRETE_PATH: &str = "/etc/prime-discrete";
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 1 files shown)


jackpot51 (approved):

jacobgkau:
This isn't currently in a staging repo even though it shows as built, may need to test by building locally.

13r0ck:
This can't be approved until it can be pulled from a staging repo, but after some preliminary testing by building locally with pop-os/nvidia-graphics-drivers#128 this works correctly.

13r0ck (approved):
Tested on Gaze16 and Addw2 on all 3 graphics profiles w/ pop-os/nvidia-graphics-dirvers#128 . Sleep still works correctly.

crawfxrd:
This needs to be tested on a TGL-U board as well, as those are the only ones using s2idle.

13r0ck:
On Kudu6, the system fails to start after a suspend

13r0ck (changes requested):
On Kudu6, the system fails to start after a suspend

crawfxrd:
Is there a SATA drive still in it? If so, it must be removed.
If not, are there any nvidia or amdgpu errors in journalctl for that boot?

And what is the output of:
```
cat /proc/driver/nvidia/params
```

13r0ck:
Oh, I wasn't fully up to date on the kudu6/sata issues. Switching to nvme fixed the issue. Let me finish some testing, then this will probably be approved.

13r0ck (approved):
Kudu6 has no regressions on 21.10, 21.04, 20.04

There is an issue with the kudu6 on 20.04 when starting from suspend. CLI says
`Can't reactivate KBD Port`, then proceeds to gnome shell. But that is not a regression. It exists without this PR

Issue #291, opened 2021-11-10
Remove dependency on ubuntu-drivers-common


crawfxrd:
We rely on gpu-manager for configuring the X server and power management. Just do it ourselves and disable `gpu-manager.service`.

- Power management: Write a value to the sysfs file `/sys/bus/pci/devices/0000:01:00.0/power/control`.
  - #297
- Configuring X: Add/Remove an X config file before the display manager starts, depending on the mode set.
  - #298

crawfxrd:
I think this is done? Would just need to drop u-d-c from the control file.

Need to test if everything works if `gpu-manager` is disabled.

PR #297, opened 2021-12-09
Configure runtimepm in sysfs

crawfxrd:
Set `power/control` for the dGPU based on the graphics mode.

Don't remove the GPU from the bus and rely on RTD3 to put it in D3.

HACK: Defer setting to "auto" for 5 seconds using a thread to avoid a bug that causes the NVIDIA driver to crash and lock up the system. May not always work. (Fixing requires system76/coreboot#98.)

Requires: pop-os/ubuntu-drivers-common#15
Ref: #291

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -34,7 +34,7 @@ use crate::{
     err_str,
     errors::ProfileError,
     fan::FanDaemon,
-    graphics::Graphics,
+    graphics::{Graphics, GraphicsMode},
     hid_backlight,
     hotplug::{Detect, HotPlugDetect},
     kernel_parameters::{KernelParameter, NmiWatchdog},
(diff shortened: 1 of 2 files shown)


jackpot51 (dismissed):

jacobgkau:
I have oryp7 to test with now, and this does not seem to fix compute mode on oryp7. When booting in compute mode, I am still seeing freezes on boot and/or login (doesn't happen every time).

I was also wondering if it would be a problem that this system76-power PR is for all releases while the corresponding ubuntu-drivers-common is only for Impish, although I can't check Focal for differences until it's at least working on Impish.

crawfxrd:
Once it working on impish I can make PRs for the other u-d-c branches.

I could disable runpm in `compute` mode as well. It would mean increased power usage...but that's pretty much expected for the mode's intended use case.

jackpot51 (dismissed):

jackpot51 (approved):

jackpot51 (approved):

jacobgkau:
On oryp7, I recreated the compute mode freezing bug without this branch on an up-to-date unencrypted install; after adding the `power-control` and `no-sysfs-runpm` branches, the freezing seems to have stopped.

I'm moving back to full regression testing on this. Can we get either a focal ubuntu-drivers-common PR or a master_focal branch of system76-power?

crawfxrd:
Made the PR for focal, including updating to `1:0.9.0~0.20.04.5` to match the Ubuntu release.

crawfxrd:
> If I turn off the GPU with system76-power graphics power off, it won't turn back on again.

That is known: https://github.com/system76/firmware-open/issues/122

jacobgkau (approved):
> That is known: [system76/firmware-open#122](https://github.com/system76/firmware-open/issues/122)

Ok. I've also had a number of times on oryp7 in both compute and NVIDIA mode where `nvidia-smi` shows `No devices were found`, with dmesg showing `nvidia 0000:01:00.0: can't shange power state from D3cold to D0 (config space inaccessible)`. That was a symptom mentioned in [this issue](https://github.com/system76/firmware-open/issues/160). For compute mode, this is still better than the freezing that's being fixed. For NVIDIA mode, I can consistently recreate it by powering off and back on 2-5 times, but it also happens on the currently-released system76-power/ubuntu-drivers-common, so it's not a regression.

Completed regression testing on both focal and impish using oryp8:
[system76-power-pr297-impish.txt](https://github.com/pop-os/system76-power/files/8353571/system76-power-pr297-impish.txt)
[system76-power-pr297-focal.txt](https://github.com/pop-os/system76-power/files/8353570/system76-power-pr297-focal.txt)

PR #298, opened 2021-12-10
Configure Xserver when switching modes

crawfxrd:
Using discrete graphics for display rendering requires setting the "PrimaryGPU" option for X.

Requires:
- pop-os/ubuntu-drivers-common#16
- pop-os/ubuntu-drivers-common#21


jackpot51 (dismissed):

jackpot51 (dismissed):

n3m0-22:
There are conflicts with this PR that need to be resolved.

n3m0-22:
The above results were the same for `21.10`.

n3m0-22:
Consistent results in `20.04`.

n3m0-22 (approved):

PR #302, opened 2022-01-04
add 'must_use' attributes to method returning 'Self'

danieleades:


jacobgkau (approved):
Change looks trivial and makes sense to me based on the [Rust `must_use` attribute documentation](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute).

Regression testing passed: [system76-power-pr302.txt](https://github.com/pop-os/system76-power/files/7953302/system76-power-pr302.txt)

Absolucy (approved):
Seems alright to me - trivial and passes regression testing, as previously mentioned.

PR #303, opened 2022-01-04
Remove use of rustfmt::skip

danieleades:


jacobgkau (approved):
The Rustfmt check is still passing, and https://github.com/pop-os/system76-power/issues/267 is still fixed (see https://github.com/pop-os/system76-power/pull/290).

jacobgkau (approved):
Confirmed on darp10-b this is still working as expected for single-color keyboard systems (using https://github.com/pop-os/system76-power/pull/290 as a reference).

crawfxrd (approved):

PR #309, opened 2022-02-24
daemon: Don't remove GPU from the bus

crawfxrd:
Requires: #297, for configuring sysfs runtimepm in system76-power instead of u-d-c.


crawfxrd:
Handled in #297.

PR #310, opened 2022-02-25
feat: Achieve higher CPU turbo boosts with cpufreq adjustments

mmstick:
I'm able to achieve higher turbo boosts on my gaze15 with this change with lightly-threaded tasks.

This should help with systems which are throttling CPU performance based on the cpufreq scaling governor.


leviport:
Just an update on testing progress so far: Nvidia galp5 still switches all graphics modes and power profiles appropriately with this update. I'm seeing only a tiny difference in CPU boost max when stressing a single core with `stress-ng`, but I'm also not seeing any issues. 

I'm going to check for regressions on a couple more machines, but this seems fine so far.

leviport (approved):
I'm not seeing dramatic boost improvements, but I'm also not seeing any regressions. Tested on galp5, kudu6, and gaze16-3050.

jackpot51 (approved):

PR #312, opened 2022-03-01
improv: Use clap 3 derive macro for argument parsing

ids1024:
Using the derive macro seems generally better to add some static verification the argument parsing code.

This seems to be working like the previous version, though `system76-power graphics --help` seems to have sub-commands and their descriptions on seperate lines, when the weren't before. I guess something in `clap` changed related to that...


src/client.rs
```diff
@@ -202,28 +211,28 @@ pub fn client(subcommand: &str, matches: &ArgMatches) -> Result<(), String> {
...
                 println!("{}", client.get_graphics()?);
                 Ok(())
             }
         },
-        "charge-thresholds" => {
+        Args::ChargeThresholds { profile, list_profiles, thresholds } => {
             let profiles = client.get_charge_profiles()?;
 
-            if let Some(mut thresholds) = matches.values_of("thresholds") {
+            if thresholds.len() != 0 {
```
Absolucy, line 222:
```suggestion
            if !thresholds.is_empty() {
```

Absolucy (changes requested):
Looks good to me, other than the one clippy lint that's easy to resolve.

Clap 3's derive macro results in nicer code, which is always nicer to have :)

src/client.rs
ids1024: Updated. I should probably run `clippy` more regularly.

Absolucy (changes requested):
Unsure why GitHub says the Jenkins build is failing, considering that the builds seem to have succeeded on Jenkins.

Should be ready after a `cargo fmt`.

XV-02 (approved):
Seems act as expected, with no obvious regression issues on desktop or laptop platforms, Looks largely good to go.

jackpot51 (approved):
This looks good to me now

jackpot51:
@ids1024 it looks like there are conflicts that will need to be resolved.

jackpot51 (approved):

Absolucy (approved):

XV-02 (approved):
[Surreptitiously copy/pastes old review after testing everything again...]
Seems to act as expected, with no obvious regression issues on desktop or laptop platforms, Looks good to go.

PR #315, opened 2022-03-04
fix: Use powersave on balanced profile

mmstick:
Fixes [this Reddit issue](https://www.reddit.com/r/pop_os/comments/t67y6i/weird_fan_behavior_with_oryx_pro_after_upgrades/)

### DIFF
--- a/src/cpufreq.rs
+++ b/src/cpufreq.rs
@@ -5,30 +5,18 @@ use crate::util::write_value;
 use concat_in_place::strcat;
 use std::fs;
 
-pub fn powersave() {
+pub fn set(mut governor: &str, max_percent: u8) {
     if let Some(driver) = scaling_driver(0) {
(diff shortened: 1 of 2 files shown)


jackpot51 (approved):

leviport (approved):
Before the update, my oryp6 ran at full speed all the time while on Balanced:


After this update, scaling is working as expected:


I'm still able to use all power modes and graphics profiles on 21.10 and 20.04 on oyrp6 and gaze16.

I did notice something a bit odd though. Switching to Performance still gets me full speeds:


but stressing a single core in Balanced mode scales all of my cores up to ~3600MHz


It seems that the 11th gen CPU in the gaze16 is a lot smarter about scaling only the core that is under heavy load. This may just be a difference in 11th gen and 10th gen though, and I can still get full speeds on Performance mode if I want them, so I'm approving.

PR #323, opened 2022-03-24
hotplug: Add oryp9 detection

crawfxrd:


crawfxrd:
mDP is actually connected to iGPU (TCP3) on gaze17-3050, so no detection is needed.

XV-02 (approved):
Works as desired. Approving.

13r0ck (approved):

Issue #325, opened 2022-03-30
GPU fan runs nonstop after update to version 1.1.20~1648241921~21.10~1e1599a


n3m0-22:
**System:**
NAME="Pop!_OS"
VERSION="21.10"
```oryp4``` 

``` 
**Software:** 
Installed: 1.1.20~1648241921~21.10~1e1599a
Candidate: 1.1.20~1648241921~21.10~1e1599a
Version table:
*** 1.1.20~1648241921~21.10~1e1599a 1001
     1001 http://apt.pop-os.org/release impish/main amd64 Packages
      100 /var/lib/dpkg/status
```

**Bug:**
Upgrading ```system76-power``` on the orxy pro 4 from version ```1.1.20~1647360548~21.10~815500d``` to the current release causes the GPU fan to run constantly no matter the temperature or CPU load. This happens under all GPU settings.

**Expected behavior:**
Normally with graphics set to integrated the GPU fan only turns on for a few seconds on occasion as pictured below

![expected fan behavior](https://user-images.githubusercontent.com/11048214/160887787-3268d7f1-539d-4eb3-81f8-41559ef9fdd3.png)

**Actual behavior:**
After installing version ```1.1.20~1648241921~21.10~1e1599a```
![fan after upgrade system76-power](https://user-images.githubusercontent.com/11048214/160888211-2306dd1b-9cb7-4c1a-b2a1-634efef8dd8b.png)

**Other Notes:**
Using ```Fn + 1``` will only reset it for a short time before it starts up again

![tmp fan reset fn plus 1](https://user-images.githubusercontent.com/11048214/160888466-c8e82088-3ccf-4fce-867d-bb0f93a7494e.png)

**Steps to Replicate:**
*Note so far this only happens on the ```oryp4```*

1. Fresh install of ```pop-os_21.10_amd64_nvidia_7```
2. Set Graphics mode to Integrated and restart then run the following
3. ```sudo apt-mark hold system76-power```
4. ```sudo apt update```
5. ```sudo apt upgrade```
6. Use psensor or lm-sensors to inspect GPU fan speeds
7. ```sudo apt-mark unhold system76-power```
8. ```sudo apt upgrade```
9. Now with the newest version of system76-power installed compare the fan speed

crawfxrd:
Does reverting 1e1599a fix it?

n3m0-22:
Yup that fixed it! Thank you.

crawfxrd:
[...]

PR #326, opened 2022-03-30
Disable dGPU in integrated mode if it doesn't support runtimepm

crawfxrd:
If a device does not support runtime power management, then remove it from the bus when in integrated mode.

Fixes: 1e1599a9dd12 ("daemon: Always enable GPU power")
Resolves: #325

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -233,11 +233,11 @@ pub async fn daemon() -> Result<(), String> {
         None
     };
 
-    log::info!("Setting graphics power");
-    match daemon.set_graphics_power(true) {
+    log::info!("Setting automatic graphics power");
+    match daemon.auto_graphics_power() {
         Ok(()) => (),
         Err(err) => {
-            log::warn!("Failed to set graphics power: {}", err);
+            log::warn!("Failed to set automatic graphics power: {}", err);
         }
     }
 
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -521,8 +521,12 @@ impl Graphics {
     }
 
     pub fn auto_power(&self) -> Result<(), GraphicsDeviceError> {
+        // Only disable power if in integrated mode and the device does not
+        // support runtime power management.
         let vendor = self.get_vendor()?;
-        self.set_power(vendor != GraphicsMode::Integrated)
+        let power = vendor != GraphicsMode::Integrated || self.gpu_supports_runtimepm()?;
+
+        self.set_power(power)
     }
 
     fn switchable_or_fail(&self) -> Result<(), GraphicsDeviceError> {
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -8,6 +8,7 @@ use std::{
     fs,
     io::{self, Write},
     iter::FromIterator,
+    path,
     process::{self, ExitStatus},
 };
 use sysfs_class::{PciDevice, SysClass};
@@ -113,15 +114,18 @@ pub enum GraphicsDeviceError {
 
 pub struct GraphicsDevice {
     id:        String,
+    devid:     u16,
     functions: Vec<PciDevice>,
 }
 
 impl GraphicsDevice {
-    pub fn new(id: String, functions: Vec<PciDevice>) -> GraphicsDevice {
-        GraphicsDevice { id, functions }
+    pub fn new(id: String, devid: u16, functions: Vec<PciDevice>) -> GraphicsDevice {
+        GraphicsDevice { id, devid, functions }
     }
 
     pub fn exists(&self) -> bool { self.functions.iter().any(|func| func.path().exists()) }
+
+    pub fn device(&self) -> u16 { self.devid }
 
     pub unsafe fn unbind(&self) -> Result<(), GraphicsDeviceError> {
         for func in &self.functions {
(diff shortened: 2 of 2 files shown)


jackpot51 (dismissed):

n3m0-22 (commented):
```21.10``` I have tested this on the ```oryp4```:

```
apt policy system76-power 
system76-power:
  Installed: 1.1.20~1648670246~21.10~77def87
  Candidate: 1.1.20~1648670246~21.10~77def87
  Version table:
 *** 1.1.20~1648670246~21.10~77def87 1002
       1002 http://apt.pop-os.org/staging/missing-runpm impish/main amd64 Packages
        100 /var/lib/dpkg/status
     1.1.20~1648241921~21.10~1e1599a 1001
       1001 http://apt.pop-os.org/release impish/main amd64 Packages

```

with the ```510``` and ```470``` drivers in both in integrated and hybrid modes and the GPU fan is always running at around 4300rpm.

Refering back to the orginal issue https://github.com/pop-os/system76-power/issues/325 the output of ```/sys/bus/pci/devices/0000:01:00.0/power/control``` in in integrated mode is now ```on``` and ```auto``` in hybrid mode.

Since I was also using this https://github.com/pop-os/system76-power/pull/329 to test the ```470``` issue earlier I also tried it here, but the problem was the same.

**Comment by crawfxrd** [comment]
Right...I query the _loaded_ drivers to get version, because NVIDIA puts the version in the folder name at `/usr/share/doc/nvidia-driver-<VERSION>`. So the check for runtimepm will fail.

**Review by jackpot51 — APPROVED** [review]

**Comment by n3m0-22** [comment]
#### On 21.10  oryp4 this seem to be working now with the 510 drivers in Integrated mode, but the GPU fan still runs nonstop with the 470 drivers in Integrated mode.

```
apt policy system76-power 
system76-power:
  Installed: 1.1.20~1649700305~21.10~e5a37b4
  Candidate: 1.1.20~1649700305~21.10~e5a37b4
  Version table:
 *** 1.1.20~1649700305~21.10~e5a37b4 1002
       1002 http://apt.pop-os.org/staging/missing-runpm impish/main amd64 Packages
        100 /var/lib/dpkg/status
     1.1.20~1648241921~21.10~1e1599a 1001
       1001 http://apt.pop-os.org/release impish/main amd64 Packages
```

**510.54**

510 54_missing-runpm_fan_integrated_mode

**510.60.02**

510 60 02_missing-runpm_fan_integrated_mode

**470.86**

470 86_missing-runpm_fan_integrated_mode

**470.103.01**

470 103 01_missing-runpm_fan_integrated_mode

**Comment by crawfxrd** [comment]
What's the journal say?

```
systemctl -b -u system76-power
```

**Comment by n3m0-22** [comment]
That just gives me ```systemctl: invalid option -- 'b'```.

**Comment by crawfxrd** [comment]
Sorry, `journalctl`, not `systemctl`.

**Comment by Brikaa** [comment]
I have a question: since the integrated mode blacklists the Nvidia drivers causing the device to be driverless, how will devices that support `runtimepm` power manage themselves in integrated mode? So why isn't removing the device from PCI the default behavior for all cards in integrated mode?

**Comment by crawfxrd** [comment]
Removing the device from the bus was the previous default. But if a device supports RTD3 then it *shouldn't* be required. The kernel will suspend the device when it's not used.

**Comment by Brikaa** [comment]
So the drivers don't need to be loaded in order for the kernel to suspend the device?

**Comment by Brikaa** [comment]
I mean, isn't power management the driver's responsibility? And in this case, it is blacklisted.

**Comment by n3m0-22** [comment]
@crawfxrd In testing `22.04` on the `orpy4` this problem still persists. I came up with a temporary solution that might help fix this. Running `system76-power graphics power off` stopped the fan issue, so I made to following.

```
[Unit]
Description=Stop GPU fan from running non-stop in Integrated Graphics mode

[Service]
ExecStartPre=/bin/sleep 30
ExecStart=/bin/bash /usr/bin/gpu-fan-fix.sh

[Install]
WantedBy=multi-user.target
```

```
#!/bin/bash

if [[ $(system76-power graphics) = 'integrated' ]]; then
    if [[ $(system76-power graphics power) = 'on (discrete)' ]]; then
        system76-power graphics power off
   fi
fi
```

jackpot51 (approved):

n3m0-22 (approved):
This fixes the fan issues in Integrated mode.

### CLI

- [x] Power profiles can be queried and set
- [x] Laptop with switchable graphics:
    - [x] Switchable graphics can be queried
        - Command returns `daemon returned an error message: "The name com.system76.PowerDaemon was not provided by any .service files` on a non-switchable system
        - Command returns `switchable` on a laptop with switchable graphics
    - [x] Switching from Integrated to NVIDIA
    - [x] Switching from Integrated to Hybrid
    - [x] Switching from Integrated to Compute
    - [x] Switching from NVIDIA to Integrated
    - [x] Switching from NVIDIA to Hybrid
    - [x] Switching from NVIDIA to Compute
    - [x] Switching from Hybrid to Integrated
    - [x] Switching from Hybrid to NVIDIA
    - [x] Switching from Hybrid to Compute
    - [x] Switching from Compute to Integrated
    - [x] Switching from Compute to NVIDIA
    - [x] Switching from Compute to Hybrid
    - [x] Discrete graphics power state can be queried and set


### GNOME Shell

- [x] Test that the power profile can be switched, and that the dots are correct
- [x] Test that any power profile change from the CLI is reflected in the extension
- [x] When switching to balanced, with screen brightness maxed, screen brightness drops to 50%
- [x] When switching to battery, with screen brightness maxed, screen brightness drops to 10%
- [x] When switching to balanced, with screen brightness minimized, screen brightness does not change
- [x] When restarting the daemon, and the daemon defaults to a balanced profile, the brightness should not change
- [x] When restarting the system, screen brightness should be the same as before
- [x] Laptop with switchable graphics:
    - [x] Switching from Integrated to NVIDIA
    - [x] Switching from Integrated to Hybrid
    - [x] Switching from Integrated to Compute
    - [x] Switching from NVIDIA to Integrated
    - [x] Switching from NVIDIA to Hybrid
    - [x] Switching from NVIDIA to Compute
    - [x] Switching from Hybrid to Integrated
    - [x] Switching from Hybrid to NVIDIA
    - [x] Switching from Hybrid to Compute
    - [x] Switching from Compute to Integrated
    - [x] Switching from Compute to NVIDIA
    - [x] Switching from Compute to Hybrid
    - [x] Test that switchable graphics changes from the CLI are reflected in the extension

crawfxrd:
@n3m0-22 did it actually work? I didn't change anything, just rebased. Do you have the workaround still applied?

n3m0-22:
Yes I completely removed the service and script I created then applied this. It had worked before on 21.10. This weekend was the first time I had a chance to fully test 22.04 on the `orpy4`. Since this was behind the current version I made the workaround as a temporary fix. I'm running the same tests now on https://github.com/pop-os/system76-power/pull/336. So far that's working too, and the `daemon returned an error message: "The name com.system76.PowerDaemon was not provided by any .service files` on a non switchable showed `not switchable` this time. I'll post what my findings there as soon as I'm finished.

crawfxrd:
If this works then #336 isn't needed.

n3m0-22:
Either way I tested both, and both solve the fan issue on the `orpy4`. The one difference being the CLI error on a non-switchable system with this one.

n3m0-22:
The graphs I posted on the other PR match the GPU fan performance change as with this PR.

PR #329, opened 2022-04-07
graphics: Disable runpm for compute mode

crawfxrd:
We still frequently see the power issue when running in compute mode. This increases power usage in compute mode, but this is acceptable as the intended use case for compute mode is for GPU intense workloads and should be used while on AC power.


jackpot51 (dismissed):

crawfxrd:
The problem will be fixed when a new firmware release is made with the coreboot driver rewrite.

PR #333, opened 2022-04-18
chore: Add system76-power.service as an alias

mmstick:
Installs the service with both names so they're addressable either way.

### DIFF
--- a/data/com.system76.PowerDaemon.service
+++ b/data/com.system76.PowerDaemon.service
@@ -9,3 +9,4 @@ BusName=com.system76.PowerDaemon
 
 [Install]
 WantedBy=multi-user.target
+Alias=system76-power.service
--- a/debian/system76-power.com.system76.PowerDaemon.service
+++ b/debian/system76-power.com.system76.PowerDaemon.service
@@ -9,3 +9,4 @@ BusName=com.system76.PowerDaemon
 
 [Install]
 WantedBy=multi-user.target
+Alias=system76-power.service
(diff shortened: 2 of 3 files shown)


bflanagin (dismissed):
Disregard approval. It does work as expected but there are a few tests that still need to be checked after the change.

jackpot51 (approved):

jacobgkau:
I'm not seeing this work.
```
system76@pop-os:~$ apt policy system76-power
system76-power:
  Installed: 1.1.20~1651624475~22.04~8b44825
  Candidate: 1.1.20~1651624475~22.04~8b44825
  Version table:
 *** 1.1.20~1651624475~22.04~8b44825 1002
       1002 http://apt.pop-os.org/staging/systemd jammy/main amd64 Packages
        100 /var/lib/dpkg/status
     1.1.20~1651262978~22.04~31a639b 1001
       1001 http://apt.pop-os.org/release jammy/main amd64 Packages
system76@pop-os:~$ systemctl status system76-power.service 
Unit system76-power.service could not be found.
system76@pop-os:~$ sudo systemctl stop system76-power
Failed to stop system76-power.service: Unit system76-power.service not loaded.
system76@pop-os:~$ 
```

This is supposed to allow using `system76-power.service` in commands, right? It actually tab completes in the `status` command, but then says it couldn't be found. The `stop` command doesn't tab complete the name, and it says it's not running even though `systemctl status com.system76.PowerDaemon.service` says that it is.

jacobgkau (approved):
The alias is working after the last change.

13r0ck (approved):

PR #335, opened 2022-04-25
Correctly apply changes to all CPU cores

lfom:
For loops in Rust do not include the last value without adding an equal signal to it.

This can be tested with:
```
watch cat /sys/devices/system/cpu/cpu[0-9]*/cpufreq/scaling_governor
```

### DIFF
--- a/src/cpufreq.rs
+++ b/src/cpufreq.rs
@@ -26,7 +26,7 @@ pub fn set(profile: Profile, max_percent: u8) {
             let max = max * max_percent.min(100) as usize / 100;
             eprintln!("setting {} with max {}", governor, max);
 
-            for cpu in 0..cpus {
+            for cpu in 0..=cpus {
                 set_frequency_minimum(cpu, min);
                 set_frequency_maximum(cpu, max);
                 set_governor(cpu, governor);


src/cpufreq.rs
```diff
@@ -26,7 +26,7 @@ pub fn set(profile: Profile, max_percent: u8) {
             let max = max * max_percent.min(100) as usize / 100;
             eprintln!("setting {} with max {}", governor, max);
 
-            for cpu in 0..cpus {
+            for cpu in 0..=cpus {
```
jackpot51, line 29: The total number of CPUs should not be used, if you have 8 CPUs then you have CPU 0, 1, 2, 3, 4, 5, 6, and 7. Can you describe the issue you are seeing?

lfom: The function that gets the total number of CPUs actually gets the highest index, starting from 0, as you said. So, in my case, the highest number is 7 (cpus = 7), and when applying the changes to system76-power the cores from 0 to 6 change, but not the last one (7). Adding the equal so the loop actually goes to 7 fixes the problem.

jackpot51: I see, you are right, it will return the number of CPUs - 1

jackpot51 (approved):

XV-02 (approved):
No regressions, seems to do what it says on the tin. I reckon it's good to go!

PR #336, opened 2022-05-02
Revert "daemon: Always enable GPU power" 

crawfxrd:
oryp4 doesn't behave correctly with the GPU enabled in integrated mode because it doesn't support RTD3.

Revert the change that left the GPU always enabled, and cherry-pick the change to get supported features while in integrated mode.

Fixes: #325

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -233,11 +233,11 @@ pub async fn daemon() -> Result<(), String> {
         None
     };
 
-    log::info!("Setting graphics power");
-    match daemon.set_graphics_power(true) {
+    log::info!("Setting automatic graphics power");
+    match daemon.auto_graphics_power() {
         Ok(()) => (),
         Err(err) => {
-            log::warn!("Failed to set graphics power: {}", err);
+            log::warn!("Failed to set automatic graphics power: {}", err);
         }
     }
 
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -8,6 +8,7 @@ use std::{
     fs,
     io::{self, Write},
     iter::FromIterator,
+    path,
     process::{self, ExitStatus},
 };
 use sysfs_class::{PciDevice, SysClass};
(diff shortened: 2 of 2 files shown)


jackpot51 (approved):

n3m0-22 (approved):
Fan issue is still fixed with this same as with https://github.com/pop-os/system76-power/pull/326, but there is no more error on a non-switchable system. The GPU fan still kicks in for a couple seconds every 3 minutes or so while working. That is in Integrated mode. The fans run normally in the NVIDIA modes. In Integrated mode the output of `system76-power graphics power` is `off (discrete)`.  These are the same results I had with the workaround.

With this PR:
Screenshot from 2022-05-02 17-13-09
 
Without:
Screenshot from 2022-05-01 11-53-37

All of the following tests passed
### CLI

- [x] Power profiles can be queried and set
- [x] Laptop with switchable graphics:
    - [x] Switchable graphics can be queried
        - Command returns `not switchable` on a non-switchable system
        - Command returns `switchable` on a laptop with switchable graphics
    - [x] Switching from Integrated to NVIDIA
    - [x] Switching from Integrated to Hybrid
    - [x] Switching from Integrated to Compute
    - [x] Switching from NVIDIA to Integrated
    - [x] Switching from NVIDIA to Hybrid
    - [x] Switching from NVIDIA to Compute
    - [x] Switching from Hybrid to Integrated
    - [x] Switching from Hybrid to NVIDIA
    - [x] Switching from Hybrid to Compute
    - [x] Switching from Compute to Integrated
    - [x] Switching from Compute to NVIDIA
    - [x] Switching from Compute to Hybrid
    - [x] Discrete graphics power state can be queried and set


### GNOME Shell

- [x] Test that the power profile can be switched, and that the dots are correct
- [x] Test that any power profile change from the CLI is reflected in the extension
- [x] When switching to balanced, with screen brightness maxed, screen brightness drops to 50%
- [x] When switching to battery, with screen brightness maxed, screen brightness drops to 10%
- [x] When switching to balanced, with screen brightness minimized, screen brightness does not change
- [x] When restarting the daemon, and the daemon defaults to a balanced profile, the brightness should not change
- [x] When restarting the system, screen brightness should be the same as before
- [x] Laptop with switchable graphics:
    - [x] Switching from Integrated to NVIDIA
    - [x] Switching from Integrated to Hybrid
    - [x] Switching from Integrated to Compute
    - [x] Switching from NVIDIA to Integrated
    - [x] Switching from NVIDIA to Hybrid
    - [x] Switching from NVIDIA to Compute
    - [x] Switching from Hybrid to Integrated
    - [x] Switching from Hybrid to NVIDIA
    - [x] Switching from Hybrid to Compute
    - [x] Switching from Compute to Integrated
    - [x] Switching from Compute to NVIDIA
    - [x] Switching from Compute to Hybrid
    - [x] Test that switchable graphics changes from the CLI are reflected in the extension

PR #337, opened 2022-05-04
graphics: Assume RTD3 for NVIDIA only

crawfxrd:
Fixes running daemon on systems with AMD GPUs and without NVIDIA GPUs.


jackpot51 (approved):

XV-02 (approved):
Successfully resolves an issue in current master which resulted in `system76-power` crashing on systems without Nvidia dGPUs. Doesn't evidence any regressions of impediments to correct function of `system76-power`. Good to go.

PR #361, opened 2022-09-16
perf: Reduce AMD Hotplug detection CPU usage with memmap

### LINKED ISSUE #360: CPU usage increases over time and eventually causes audio to stutter
 Pop!\_OS 22.04
 
 
Presumably `system76-power`, version unknown

**Issue/Bug Description:**

https://twitter.com/emptythevoid/status/1570150727300923393

Not entirely clear, but presumably the machine just has to be left running with music playing/looping in the background for a while?

**Expected behavior:**
system76-power should not hog CPU, and audio playback should not stutter

**Other Notes:**
I assume this is on a kudu6, judging by the core count in `powertop` in the video.

mmstick:
May fix #360 

Needs testing to ensure display hotplug detection still works with AMD integrated graphics laptops.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -163,6 +163,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "errno"
+version = "0.2.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f639046355ee4f37944e44f60642c6f3a7efa3cf6b78c78a0d989a8ce6c396a1"
+dependencies = [
+ "errno-dragonfly",
+ "libc",
+ "winapi",
+]
+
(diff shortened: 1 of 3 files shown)


XV-02:
This seems to be breaking hotplug detection on Kudu. 
Without this PR, in integrated mode, the attempt to attach a monitor - in this case over mini-displayport - is detected and the prompt to switch to hybrid mode is presented. 
With this PR, the prompt is not present.
When I look through the output from `journalctl` I noticed that `system76-power` did not report `[INFO] HotPlugDetect 2`

mmstick:
It doesn't seem to fix the performance issue, so a different approach will be necessary

Issue #370, opened 2022-11-17
GTX 1030 and RTX 4090 being disabled during Pop! setup


jacobktm:
**Distribution (run `cat /etc/os-release`):**
 
 NAME="Pop!_OS"
VERSION="22.04 LTS"
ID=pop
ID_LIKE="ubuntu debian"
PRETTY_NAME="Pop!_OS 22.04 LTS"
VERSION_ID="22.04"
HOME_URL="https://pop.system76.com"
SUPPORT_URL="https://support.system76.com"
BUG_REPORT_URL="https://github.com/pop-os/pop/issues"
PRIVACY_POLICY_URL="https://system76.com/privacy"
VERSION_CODENAME=jammy
UBUNTU_CODENAME=jammy
LOGO=distributor-logo-pop-os
 
**Related Application and/or Package Version (run `apt policy $PACKAGE NAME`):**

system76-power:
  Installed: 1.1.24\~1667870429\~22.04\~e0a6fb8
  Candidate: 1.1.24\~1667870429\~22.04\~e0a6fb8
  Version table:
 *** 1.1.24\~1667870429\~22.04\~e0a6fb8 1001
       1001 http://apt.pop-os.org/release jammy/main amd64 Packages
        100 /var/lib/dpkg/status

**Issue/Bug Description:**

While installing Pop! OS 22.04 with GTX 1030 or RTX 4090 installed with iGPU enabled the discrete graphics card is disabled.

I think this commit is potentially the cause of this issue: https://github.com/pop-os/system76-power/commit/36db16adad2af58029bb1b4926d3adc7931d0e83

**Steps to reproduce (if you know):**

1) Install a GTX 1030 or RTX 4090 into the system with display out plugged into discrete GPU (make sure integrated graphics is enabled in BIOS)
2) install Pop! OS 22.04
3) on reboot after installing system hangs until display out plugged in to integrated graphics

**Expected behavior:**

**Other Notes:**

jacobktm:
I kept digging in to this issue and I believe the actual root cause is here: https://github.com/pop-os/system76-power/issues/370

jacobktm:
I closed the wrong issue by mistake

jacobktm:
fixed by https://github.com/pop-os/system76-power/pull/373

PR #372, opened 2022-11-28
Add check for thelio to get_default_graphics

jacobktm:
Checks to see if the system is a thelio in get_default_graphics and enables hybrid graphics if true. This fixes [issue #370 ](https://github.com/pop-os/system76-power/issues/370)

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -392,6 +392,7 @@ impl Graphics {
             .map_err(GraphicsDeviceError::SysFs)
             .map(|s| s.trim().to_string())?;
         let blacklisted = DEFAULT_INTEGRATED.contains(&product.as_str());
+        let thelio = product.contains("thelio");
 
         let runtimepm = self.gpu_supports_runtimepm().unwrap_or_default();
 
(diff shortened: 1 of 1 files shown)


13r0ck (dismissed):

crawfxrd (changes requested):
Desktops should be defaulting to discrete graphics.

Does NVIDIA support hybrid graphics in a desktop configuration? The Design Guides heavily imply it is for laptops only, but I can't find anything explicitly saying it's not supported like they do for RTD3.

jacobktm:
Yes, it does, as the function is currently it's been setting all of the desktops up with hybrid graphics. I can put that test on the discrete branch and I'm pretty sure that will work too.

crawfxrd:
Then it is wrong. Desktops should be set up with discrete graphics. They should also be returning `not switchable` when queried.

jacobgkau:
> Desktops should be set up with discrete graphics. They should also be returning `not switchable` when queried.

E.g. compute mode is a valid configuration for a Thelio (to output display from integrated graphics while using a dedicated GPU for compute.)

src/graphics.rs
```diff
@@ -392,6 +392,7 @@ impl Graphics {
             .map_err(GraphicsDeviceError::SysFs)
             .map(|s| s.trim().to_string())?;
         let blacklisted = DEFAULT_INTEGRATED.contains(&product.as_str());
+        let thelio = product.contains("thelio");
```
crawfxrd, line 395: This can be generalized by checking for any desktop: `cat /sys/class/dmi/id/chassis_type` = 0x03

jacobktm:
https://github.com/pop-os/system76-power/pull/373 is a much better fix for issue #370

PR #373, opened 2022-11-28
graphics: Do not report desktops as switchable

crawfxrd:
Desktops should not report as switchable if they have a dGPU. The switchable graphics mechanisms are designed for notebook systems.

The physical selection of ports the *user* chooses when plugging in displays determines whether the iGPU or the dGPU will be used.

### Test

- Desktop reports "not switchable" when queried
- gnome-initial-setup does not show the panel about switchable graphics on desktops
- Display output works when plugged into iGPU ports
- Display output works when plugged into dGPU ports
- dGPU can be used as a compute node when displays are plugged into the iGPU ports

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -318,9 +318,18 @@ impl Graphics {
         Ok(Graphics { bus, amd, intel, nvidia, other })
     }
 
+    pub fn is_desktop(&self) -> bool {
+        let chassis = fs::read_to_string("/sys/class/dmi/id/chassis_type")
+            .map_err(GraphicsDeviceError::SysFs)
+            .unwrap_or_default();
+
+        chassis.trim() == "3"
+    }
+
     #[must_use]
     pub fn can_switch(&self) -> bool {
-        !self.nvidia.is_empty() && (!self.intel.is_empty() || !self.amd.is_empty())
+        !self.is_desktop()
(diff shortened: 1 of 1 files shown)


crawfxrd:
The existing file in `/etc/modprobe.d/` may be an issue.

jacobktm:
I'll test this out and let you know if that causes any problems

jacobktm:
After installing this on Mira B3 with an RTX 2080, when I run the command `system76-power graphics switchable` it's still reporting the graphics mode as switchable. I'm also seeing this behavior during system setup if I install system76-power from this branch in my live environment before installing Pop on a desktop it still enables hybrid graphics with GPUs that support runtimepm and disables the GPU with the GT 1030, which does not support runtimepm.

crawfxrd:
What's the output of `cat /sys/class/dmi/id/chassis_type`?

jacobktm:
> What's the output of `cat /sys/class/dmi/id/chassis_type`?

The output is "3", as expected.

crawfxrd:
It's a damn newline.

jacobktm (approved):
This works and solves [issue #370](https://github.com/pop-os/system76-power/issues/370). I also tested this on galp5 with a 3050, and it correctly detects the graphics as switchable. I confirmed that on Mira B3 each of the test conditions succeeds.

mmstick (approved):

PR #379, opened 2023-01-11
Disable turbo/boost on balanced profile when on battery

mmstick:
On systems that do not limit the power draw when on battery, the turbo and boost feature can significantly increase energy consumption, due to increasing voltage(s) in the CPU to reach the boost frequencies. Some laptops are reportedly draining battery quickly as a result, even with mundane tasks. Balanced should strive to be more energy efficient when on battery, so I think it makes sense to disable it when on battery. It also makes sense to default to enabling it when plugged into a charger.

You can get the turbo value from intel_pstate from `/sys/devices/system/cpu/intel_pstate/no_turbo`, and the boost value for AMD from `/sys/devices/system/cpu/cpufreq/boost`. May take a second or two for the event to propagate to the daemon.

I've additionally included some code refactoring improvements in the daemon:

- The HID backlight daemon no longer requires to spawn an extra thread now that it's async. Instead, it's spawned on the current_thread executor and yields between loops, or when inotify's event_stream yields.
- The hotplug detect code in the daemon's main function is now wrapped into a `hotplug::Emitter` type to make it more readable
- `cargo build --example hotplug && sudo target/debug/examples/hotplug` will run the new hotplug detect example.

In the future, we might consider disabling turbo automatically when low on battery. And at some point replacing dbus with zbus. Along with configurable settings and GUI controls further down the road.


jackpot51 (changes requested):
I don't agree with unilaterally disabling turboboost when on battery. It can leave a ton of performance on the table, and the firmware should really be determining power limits when on battery.

mmstick:
Should this be a form of configuration option then? I've been recommended that [auto-cpufreq](https://github.com/AdnanHodzic/auto-cpufreq) has this behavior by default, and it was the [main inspiration](https://foolcontrol.org/?p=3124) for creating it.

mmstick:
Or perhaps it'd make sense to have this behavior for the battery profile

jackpot51:
> Or perhaps it'd make sense to have this behavior for the battery profile

I would be okay with that.

XV-02:
In terms of system76 hardware, what systems would this apply to? 
I thought most of our current fleet did limit power draw on battery?

mmstick:
All systems. Currently, turbo is disabled by default in the battery profile, so this will enable turbo when plugged into AC, and then disable it when unplugged. Seems to be of great benefit to Ryzen laptops at the moment since a lot of firmware is limiting the frequencies to 400 MHz in the amd-pstate powersave governor when boost is turned off.

XV-02:
When I've been testing this, I've not been seeing evidence of boosted performance under the Battery Life profile while on AC.

Looking at Pang11 and Dev One as specific test cases, I'm seeing the same results from benchmarks on AC regardless of profile, and a decrease in performance on battery in the Battery Life profile (always with %70 or greater charge) with this branch installed compared to before. I'm also seeing reported clock speeds the same on AC in particular both before and after this branch was installed on both systems when running benchmarks or even just running a stress command. 

The drop in performance under battery in the Battery Life profile is ~17% based on pulling numbers from a 7zip compression benchmark using phoronix test suite. That benchmark was selected as it seems to scale well with both clockspeeds and core counts. 

I would have expected, based on the comment that the battery profile axes turbo-boost generally at current, to see this increase AC performance under Battery Life, and maintain performance under battery powered conditions?

SuspiciousCodeJester:
To optimize battery life on laptops, it is recommended to disable _**hwp_dynamic_boost**_ while on Battery Life profile, as it allows the CPU to conservatively boost it's clock rate and drains battery life faster. Reducing _**max_perf_pct**_ to 30 or 40 is optimal for mid to lower-end laptops, while a setting of 20 is recommended for lower-end laptops.

Implementing power profile switching based on AC power or battery status is a common and basic feature found in many OS's, and enabling it on Pop OS would be beneficial.

While Pop OS is developed specifically for System76 computers, enabling these features or allowing users to make these changes through the command line or settings would be beneficial non-System76 users.

Disabling **hardware pstate boost** and reducing _**max_perf_boost**_ can significantly increase laptop battery life.

PR #381, opened 2023-01-20
ci: Remove use of actions-rs

crawfxrd:
Update the toolchain file to the TOML format and specify clippy and rustfmt as required components.

Remove use of all actions-rs actions, which are unmaintained.

- None of the actions have been updated in over 2 years
- The actions-rs organization is a single user, who is no longer active on GitHub

The only thing really lost with this is the removable of clippy reporting to GitHub.

This fixes the deprecation warnings from GitHub Actions.

CC @danieleades


jackpot51 (dismissed):

XV-02:
Do these warnings:
```
warning: tool `rustfmt` is already installed, remove it from `/home/runner/.cargo/bin`, then run `rustup update` to have rustup manage this tool.
warning: tool `cargo-fmt` is already installed, remove it from `/home/runner/.cargo/bin`, then run `rustup update` to have rustup manage this tool.
```

during the `install toolchain` step of `Rustfmt` matter?

crawfxrd:
I think it's fine? I'm not sure why that warning happens with the GitHub runners. It should still be using the version specified by the override.

danieleades:
Yeah makes sense to stop using those.

I'm using https://github.com/giraffate/clippy-action these days as a linting action in my repos

XV-02 (approved):
The ci workflow looks like it is working on this branch, so I'm calling it good. 

My only hesitation are the two warnings issued during the `install toolchain` step of `Rustfmt`. I'm inclined to disregard them, as they pertain to allowing rustup to manage update installed components. Given that the ci environment builds every time, this shouldn't matter.

jackpot51 (approved):

PR #384, opened 2023-02-03
Add new models

crawfxrd:
- gaze18: All display ports are connected to the iGPU
- serw13: mDP over USB-C doesn't seem to work on proprietary firmware


XV-02 (approved):
Addw3 looks good, as does Serw13. Oryp11 and Bonw15 I don't know about, but worst case is we push the tweaks for those later, as this is needed for Addw3 launch.

jackpot51 (approved):

PR #395, opened 2023-05-02
Update with latest commit (Lunar)

jacobgkau:
Aside from wanting hardware support for these systems in 23.04, this PR is also intended to try and get the package to build, since `master_lunar` has currently failed to build (possibly due to the same transient error that the DKMS packages were seeing for a bit.)


jacobgkau:
`master_lunar` isn't protected yet, and this is merging a commit that was approved for `master`, so I went ahead and merged it.

PR #401, opened 2023-06-07
Notify in CLI if system is incompatible

13r0ck:
Print out a error message if the system is not supported. Behaviour before this is a success that looks like the power profile was changed, which is incorrect.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -89,6 +89,11 @@ impl Power for PowerClient {
         r.get1().ok_or_else(|| "return value not found".to_string())
     }
 
+    fn get_desktop(&mut self) -> Result<bool, String> {
+        let r = self.call_method::<bool>("GetDesktop", None)?;
+        r.get1().ok_or_else(|| "return value not found".to_string())
+    }
+
     fn set_graphics(&mut self, vendor: &str) -> Result<(), String> {
         println!("setting graphics to {}", vendor);
         let r = self.call_method::<&str>("SetGraphics", Some(vendor)).map(|_| ());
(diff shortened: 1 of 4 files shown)


13r0ck:
Do we want rustfmt to pass? It looks like other PRs have merged without that, so if I `cargo fmt` here it is going to change a lot of stuff

jacobgkau:
That's up to you/engineering. That check not passing isn't a regression from a QA standpoint since it's already not passing on master.

jacobgkau (changes requested):
Before, I did already get an error message that somewhat explained what the problem was if I tried to switch graphics profiles on a non-switchable system.

```
system76@pop-os:~$ system76-power graphics nvidia
setting graphics to nvidia
daemon returned an error message: "does not have switchable graphics"
system76@pop-os:~$
```

Now, I get a longer message that may be more human-readable.

```
system76@pop-os:~$ system76-power graphics nvidia

System76 Power is not supported on this device.
This means that your device is either a desktop or doesn't have both an iGPU and dGPU.
System76 Power's daemon automatically stops on systems like this. There is no need
to uninstall nor any need for further configuration.

system76@pop-os:~$ 
```

However, this new message is also showing up if I try to switch power profiles (testing on lemp10.) All laptops should work with power profiles, not just the graphics-switchable ones. This change also breaks setting charging profiles (displaying the same message.)

13r0ck:
I force pushed a change, hows it look now?

src/client.rs
```diff
@@ -173,43 +178,62 @@ pub fn client(args: &Args) -> Result<(), String> {
...
+                Some("performance") => client.performance(),
+                _ => profile(&mut client).map_err(err_str),
+            }
+        }
+        Args::Graphics { cmd } => {
+            if !client.get_switchable().unwrap_or(false) {
+                return Err(String::from(r#"
+System76 Power is not supported on this device.
+This means that your device is either a desktop or doesn't have both an iGPU and dGPU.
+System76 Power's daemon automatically stops on systems like this. There is no need
```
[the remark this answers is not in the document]
13r0ck: updated, does that look better?

src/client.rs
```diff
@@ -173,43 +178,60 @@ pub fn client(args: &Args) -> Result<(), String> {
...
+                Some("balanced") => client.balanced(),
+                Some("battery") => client.battery(),
+                Some("performance") => client.performance(),
+                _ => profile(&mut client).map_err(err_str),
+            }
+        }
+        Args::Graphics { cmd } => {
+            if !client.get_switchable().unwrap_or(false) {
+                return Err(String::from(r#"
+Graphics switching is not supported on this device, beacuse
```
jackpot51, line 198: because is spelled `beacuse`

13r0ck: oops thanks

jackpot51 (dismissed):

jacobgkau (commented):
Power profiles and charging profiles now work again on non-switchable laptops.

For desktops, do we want to consider also using the `get_desktop` method/message for charging profiles? Currently, it still returns `Not running System76 firmware with charge threshold support`.

jacobgkau (approved):
The "not switchable" string still seems pointless since it will never be displayed, but that's more engineering's discretion.

Desktop:
- `system76-power graphics` returns a new message instead of functioning like a laptop.
- `system76-power profile` returns a new message instead of functioning like a laptop.
- `system76-power charge-thresholds` returns an existing daemon error message.

I noticed in the logs that the GNOME Shell extension still thinks the "balanced" profile is being loaded when it first starts. The top-right menu doesn't include the battery/power drop-down, so not sure what its behavior would be now.

Laptop (switchable):
- `system76-power graphics` still works.
- `system76-power profile` still works.
- `system76-power charge-thresholds` still works.

Laptop (non-switchable):
- `system76-power graphics` returns a new message instead of the old daemon error message.
- `system76-power profile` still works.
- `system76-power charge-thresholds` still works.

13r0ck:
@jacobgkau 
```
system76-power charge-thresholds returns an existing daemon error message.
```
That is a good point. I pushed a fix for that. Thank you

src/client.rs
```diff
@@ -173,44 +178,68 @@ pub fn client(args: &Args) -> Result<(), String> {
...
-                println!("{}", client.get_graphics()?);
-                Ok(())
             }
-        },
+        }
         Args::ChargeThresholds { profile, list_profiles, thresholds } => {
+            if client.get_desktop()? {
+                return Err(String::from(
+                    r#"
+Charge Thresholds are not supported on desktop computers.
```
jacobgkau, line 238: Thanks. Can you un-capitalize Thresholds to match the other messages?

13r0ck: Not sure how I messed that up again... :facepalm: 
Fixed. Thank you

jacobgkau (approved):
Desktop:
- `system76-power graphics` returns a new message instead of functioning like a laptop.
- `system76-power profile` returns a new message instead of functioning like a laptop.
- `system76-power charge-thresholds` returns a new message instead of the old daemon error message.

Laptop (switchable):
- `system76-power graphics` still works.
- `system76-power profile` still works.
- `system76-power charge-thresholds` still works.

Laptop (non-switchable):
- `system76-power graphics` returns a new message instead of the old daemon error message.
- `system76-power profile` still works.
- `system76-power charge-thresholds` still works.

mmstick (approved):

jacobgkau:
I gave this a quick sanity check since it's been a while.

This is still working as expected on switchable and non-switchable laptops.

Testing in a virtual machine attempting to simulate a desktop, I found that by default, `/sys/class/dmi/id/chassis_type` was `1` (while the host laptop was `9`), which doesn't match the `3` the code is looking for to detect desktops. Looking at [DMI documentation](https://www.dmtf.org/sites/default/files/standards/documents/DSP0134_3.2.0.pdf), it appears `1` is "other" and `3` is indeed "desktop." In the virtual machine set to `1`, I was visually able to switch between power profiles using the GNOME extension, but I got errors doing it on the CLI, which is not a regression.

In order to properly simulate running this on a desktop, I cloned this branch and modified the `is_desktop` check in `graphics.rs` to look for `1` instead of `3`. After doing that, I saw similar behavior: the power profile was visually able to switch in the GUI, but attempting to switch in the CLI (or check what mode I'm in) now gives the notice that it's not supported. I do think the only reason the menu was showing up in the GUI was because VirtualBox is passing my laptop's battery info through to the VM.

We might be able to better hide the options in the GUI on desktops, but this isn't going to cause any crashes in the GUI (and it sounds like profiles weren't properly functioning on desktops anyway, from this PR's initial description).

mmstick:
Power profiles do work on desktops, yes.

PR #402, opened 2023-06-14
Update deps to fix CVE-2023-22466

13r0ck:
https://github.com/pop-os/system76-power/security/dependabot/1
https://github.com/pop-os/system76-power/security/dependabot/2 (Not a CVE, but this will resolve that dependabot alert too)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,14 +3,52 @@
 version = 3
 
 [[package]]
-name = "atty"
-version = "0.2.14"
+name = "anstream"
+version = "0.3.2"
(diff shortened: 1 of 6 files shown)


13r0ck:
I pushed a `cargo fmt` commit onto this. If we prefer not to, then I will remove that and push a commit that removes that from CI. It seems strange to me to have CI we expect to fail.

mmstick (approved):

leviport (approved):
Not seeing any regressions with `system76-power`

PR #406, opened 2023-06-30
hid_backlight: Correct colors on bonw15

jacobgkau:
This gets colors working more correctly on the bonw15. (Thanks @crawfxrd for the help with the bitmasks.)

Still need to regression test on the bonw14.

### DIFF
--- a/src/hid_backlight.rs
+++ b/src/hid_backlight.rs
@@ -11,16 +11,26 @@ fn keyboard(device: &HidDevice, brightness: u8, color: u32) -> HidResult<()> {
     let raw_brightness = (((brightness as u16) * 10 + 254) / 255) as u8;
     log::debug!("keyboard brightness {}/10 color #{:06X}", raw_brightness, color);
 
+    // Determine color channel values
+    let r = (color >> 16) as u8;
+    let mut g = (color >> 8) as u8;
+    let mut b = color as u8;
+
+    // Color correction based on model
+    let dmi_vendor = fs::read_to_string("/sys/class/dmi/id/sys_vendor").unwrap_or(String::new());
+    let dmi_model =
+        fs::read_to_string("/sys/class/dmi/id/product_version").unwrap_or(String::new());
+    match (dmi_vendor.trim(), dmi_model.trim()) {
+        ("System76", "bonw15") => {
+            g = (((g as u16) * 0x65) / 0xFF) as u8;
+            b = (((b as u16) * 0x60) / 0xFF) as u8;
+        }
+        _ => {}
+    }
+
     // Set all LED colors
     for led in 0..=255 {
-        device.send_feature_report(&[
-            0xCC,
-            0x01,
-            led,
-            (color >> 16) as u8,
-            (color >> 8) as u8,
-            color as u8,
-        ])?;
+        device.send_feature_report(&[0xCC, 0x01, led, r, g, b])?;
     }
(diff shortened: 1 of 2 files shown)


crawfxrd:
Something to try:

Swap the keyboards on bonw14 and bonw15, and flash EC firmware with `KEYBOARD` swapped.

If the problem goes away on bonw15 and appears on bonw14, then it's the keyboard itself that is the problem.

jacobgkau:
Fixed the clippy issue.

I'll try the bonw14/bonw15 keyboard swap tomorrow, as long as the lab's bonw15 isn't tied up with more pressing product development tasks. Just to be clear, we saw the blue tint on both of the bonw15 units (QA and engineering), so even if it is the keyboard itself, I'm assuming it's a quirk of the model rather than a single defective unit.

src/hid_backlight.rs
```diff
@@ -11,16 +11,31 @@ fn keyboard(device: &HidDevice, brightness: u8, color: u32) -> HidResult<()> {
...
+
+    // Set all keys to the same color, correcting color based on model
+    if model.trim() == "bonw15" {
+        for led in 0..=255 {
+            device.send_feature_report(&[
+                0xCC,
+                0x01,
+                led,
+                (color >> 16) as u8,
+                ((color >> 8) as u8) & 0x65,
```
jackpot51, line 24: I don't think it is best to apply the limit this way. It will lose more precision than to multiply by 0x65 and divide by 0xff

jackpot51: @jacobgkau I have applied the change I believe will provide better accuracy. Please see if the color looks good to you.

jackpot51:
Pushed again with clippy fixed

jacobgkau:
Answering @crawfxrd's earlier question, with no fixes applied:

- With their original keyboards, first confirmed I see the issue on bonw15 and not bonw14.
- Swapped the keyboards (without changing EC firmware); the bonw14's keyboard in the bonw15 did not have the issue, while the bonw15's keyboard in the bonw14 did have the issue.
- Flashed firmware to both machines with the EC firmware's `KEYBOARD` swapped (required for typing to work properly); there's no change in behavior, the bonw15 keyboard in bonw14 is still blue and the bonw14 keyboard in bonw15 is still white. (Didn't expect this to do anything, since `bonw15` was already using the same `KBLED` as `bonw14`.)

This confirms that it's the keyboard model and not anything else about the bonw15 (like controllers on the motherboard) causing the issue.

The bonw14 shipped with keyboard model `CVM19N23USJ430`, which has part number `6-80-X170S-012-1`. The bonw15 ships with what looks like part number `6-23-RNL7P-051 102-18K23LHA03` and model number `G1CV18K22U-COX12301130282`. (The sticker on the bown15's keyboard doesn't label what each of the numbers is.)

jacobgkau:
@jackpot51 The latest change you made to the correction code isn't working properly. It now defaults to slightly green, and cycling through the colors, some repeat and others are missing, similar to when I attempted to do math on the f32 instead of the u8. (For example, `0000FF` is completely dark while it should be blue; `FF0000` and `FF00FF` are both solid red.) This is on commit 072d50e.

jacobgkau:
Confirmed this causes no regressions on bonw14.

leviport (approved):
Also looks good on bonw15

PR #418, opened 2023-10-15
Refactor/clippy

danieleades:


mmstick (approved):

jacobgkau (approved):
<details>
<summary>Regression testing passed:</summary>

### CLI

    - [X] Command returns `not switchable` on a non-switchable system
- [X] Discrete graphics power state can be queried and set (compute mode)
    - `NVM` error attempting to remove device with non-zero usage count; not a regression.

### GNOME Shell

    - https://github.com/pop-os/gnome-shell-extension-system76-power/issues/76 still occurs, not a regression.

</details>

PR #419, opened 2023-10-15
add dependabot config

danieleades:
this PR adds dependabot for github actions only.

Could also be extended to include the projects deps, but given this a binary application dependabot can be quite noisy, so it would need some discussion about how much to clamp it down. For example configure it to run once a month, or something


jacobgkau (approved):
From what I understand, this configuration will check daily and open a PR if a newer version for an action specified in https://github.com/pop-os/system76-power/blob/master/.github/workflows/CI.yml is available (e.g. we use `actions/checkout@v3`, it will open a PR to bump the version if `actions/checkout@v4` becomes available.)

I'm not sure how necessary this is, but I don't see a QA reason to block it. The only concern I can think of would be using up Actions minutes, but [Dependabot itself is free](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/about-dependabot-security-updates), so the only additional billable usage would be running actions on the PRs being opened. It's up to engineering if it's useful/appropriate to keep the actions up-to-date.

Issue #424, opened 2024-03-16
March 12th update kills performance of Thelio Major


laniakea64:
**Distribution (run `cat /etc/os-release`):**
 
```
$ cat /etc/os-release 
PRETTY_NAME="Ubuntu 22.04.4 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.4 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy
```
 
**Related Application and/or Package Version (run `apt policy $PACKAGE NAME`):**

```
$ apt policy system76-power
system76-power:
  Installed: 1.1.26~1710180257~22.04~d7f73b2~dev
  Candidate: 1.1.26~1710180257~22.04~d7f73b2~dev
  Version table:
 *** 1.1.26~1710180257~22.04~d7f73b2~dev 1001
       1001 https://ppa.launchpadcontent.net/system76-dev/stable/ubuntu jammy/main amd64 Packages
        100 /var/lib/dpkg/status
```

**Issue/Bug Description:**

After updating to `system76-power` version `1.1.26~1710180257~22.04~d7f73b2~dev`, performance of my Thelio Major (thelio-major-b2) became severely degraded.  The effect was immediate upon installing the update.

Some investigation pointed to https://github.com/pop-os/system76-power/commit/27192d28478e42a4d7fa4c21336f9abecb900188 as the culprit: In order to get reasonable performance from this Thelio Major, I must run `system76-power profile performance` (I set up a systemd service to do this automatically).  But after that commit, it now fails:
```
$ system76-power profile performance

Power profiles are not supported on desktop computers.

```

Please reinstate support for the Performance power profile on Thelio Major, as it is critical for being able to use this system for most of its intended purposes.  Thanks.

**Steps to reproduce (if you know):**

Install `system76-power` version `1.1.26~1710180257~22.04~d7f73b2~dev` on a Thelio Major, then try to run `system76-power profile performance`.

**Expected behavior:**

Running `system76-power profile performance` should succeed and unlock the full performance of Thelio Major.

[...]

PR #425, opened 2024-03-16
Allow using Performance power profile on desktop

laniakea64:
Fixes https://github.com/pop-os/system76-power/issues/424

This patch assumes https://github.com/pop-os/system76-power/commit/27192d28478e42a4d7fa4c21336f9abecb900188 specifically intended to disable Battery power profile on desktops, since desktop computers don't have battery power like laptops.  I don't know what the power profiles do internally, so please let me know if that assumption is incorrect.

Also ran `rustfmt` on the changed area of that file, hoping to keep code style consistent with existing code.

### DIFF
--- a/src/client.rs
+++ b/src/client.rs
@@ -192,7 +192,7 @@ Battery power profile is not supported on desktop computers.
             }
             Some("performance") => client.performance(),
             _ => profile(&mut client).map_err(err_str),
-        }
+        },
         Args::Graphics { cmd } => {
             if !client.get_switchable()? {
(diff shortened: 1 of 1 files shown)


mmstick (dismissed):

mmstick:
There's no need. I will squish and rewrite the commit description on merge.

leviport (approved):
Profiles still work on laptops, and this enables switching between balanced and performance on desktops. Thanks!

mmstick (approved):

PR #426, opened 2024-03-25
Gracefully handle missing brightness_hw_changed

### LINKED ISSUE #397: system76-power daemon panic because missing file
**Distribution (run `cat /etc/os-release`):** NixOS
 
 
 
**Related Application and/or Package Version (run `apt policy $PACKAGE NAME`):** system76-power and system76-acpi

**Issue/Bug Description:** When starting system76-power on lemp12 I get the following error: ` thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/hid_backlight.rs:80:71`

Here I think the code is looking for  `sys/class/leds/system76_acpi::kbd_backlight/brightness_hw_changed`   I double checked and indeed that file does not exist. 
brightness and max_brightness does though. The keyboard led do work though.

1. Install nix os. 
2. Set: 
```services.power-profiles-daemon.enable = false;
  hardware.system76.firmware-daemon.enable = true;
  hardware.system76.kernel-modules.enable = true;
  hardware.system76.power-daemon.enable = true;
```

This installs system76_acpi and syste76_io kernel modules, the firmware updater, and system-76 power. 

```
[root@xxxx:/sys/class/leds/system76_acpi::kbd_backlight]# lsmod |grep system
system76_acpi          20480  0
battery                28672  1 system76_acpi
system76_io            16384  0
usbcore               339968  4 xhci_hcd,system76_io,btusb,xhci_pci
led_class              20480  5 snd_hda_codec_generic,sdhci,input_leds,iwlmvm,system76_acpi

[root@xxx:/sys/class/leds/system76_acpi::kbd_backlight]# system76-power --version
system76-power 1.1.23
```

**Expected behavior:**system76-power starts and does not panic.

**Other Notes:**
seems close to this: https://github.com/pop-os/system76-power/pull/290 where error handling was added for missing files.

scrufulufugus:
Prevents missing `/sys/class/leds/system76_acpi::kbd_backlight/brightness_hw_changed` from causing a crash. Tested on a `lemp9` laptop under NixOS 23.11 with system76-power 1.1.26.

Closes #397

### DIFF
--- a/src/hid_backlight.rs
+++ b/src/hid_backlight.rs
@@ -87,7 +87,9 @@ pub fn daemon() {
     let mut inotify = Inotify::init().unwrap();
     let mut watches = inotify.watches();
     watches.add(&brightness_file, WatchMask::MODIFY).unwrap();
-    watches.add(&brightness_hw_changed_file, WatchMask::MODIFY).unwrap();
+    if let Err(e) = watches.add(&brightness_hw_changed_file, WatchMask::MODIFY) {
+        log::warn!("hid_backlight: failed to watch hardware changed file: {}", e);
+    }
     if let Err(e) = watches.add(&color_file, WatchMask::MODIFY) {
         log::warn!("hid_backlight: failed to watch keyboard color: {}", e);
     }


scrufulufugus:
Also if anyone has an insight into why the NixOS kernel is missing this interface I can try to fix it on that side. It seems like upstream kernel doesn't have this issue?

mmstick (approved):

XV-02 (approved):
This is looking good. Not seeing any regressions.

PR #430, opened 2024-04-29
1.2.0: zbus daemon rewrite & org.freedesktop.UPower.PowerProfiles compat

mmstick:
- Rewrites the daemon's dbus logic with zbus
- Implements org.freedesktop.UPower.PowerProfiles for integration with software that uses it
- Provides zbus 3.15.2 crate for consumers of system76-power


jackpot51 (dismissed):

jacobgkau (changes requested):
On Pop!_OS 22.04 (in the default GNOME environment), I'm seeing the GNOME Shell extension not show any dots on the power profile. Selecting one of the profiles makes a dot appear, but does not actually switch the profile according to the CLI.

mmstick:
I will log into GNOME and compare the output dbus inspection to our extension. Did we check if the upower power profiles works for Ubuntu?

jacobgkau:
On 01a3515, I am not seeing the power profiles exposed in the GUI in Ubuntu, so I don't think they're working.

mmstick:
@jacobgkau Fixed. `com.system76.PowerDaemon` had a typo, and the `Profiles` method was missing for the upower power profiles. Tested with `powerprofilesctl`

jackpot51 (dismissed):

jacobgkau:
With 9372dc2, the dots do work as expected in the GUI Pop!_OS menu (GNOME Shell extension), and selecting profiles seems to work based on the screen brightness changing. However, the `system76-power profile` command now lists `Power Profile: ?` regardless of which power profile is currently active:

Screenshot from 2024-04-30 12-21-43

In addition, I'm still not seeing the options available in the Ubuntu GUI. (Regarding testing with `powerprofilesctl`, since that command is provided by the `power-profiles-daemon` package and `system76-power` still conflicts with that, I'm not sure if you mean it should also work with `system76-power` or if you just meant comparing the behavior.)

mmstick:
I manually compiled the latest version of power-profiles-daemon, and used powerprofilesctl with system76-power.

```
mmstick@Pop-Dev-One:~$ powerprofilesctl get
balanced
mmstick@Pop-Dev-One:~$ powerprofilesctl set performance
mmstick@Pop-Dev-One:~$ powerprofilesctl get
performance
mmstick@Pop-Dev-One:~$ powerprofilesctl list
  power-saver:

* performance:
    Degraded:   no

  balanced:
```

jackpot51 (dismissed):

mmstick:
Installing Ubuntu 24.04 to test compatibility there.

mmstick:
It would seem that gnome-shell in Ubuntu 24.04 uses the legacy name of the power profiles daemon: `net.hadess.PowerProfiles`.

jacobgkau:
In case it's important, on 16cfc5b, I noticed `system76-power` is failing to start with:

```
system76-power[3393]: unable to create system service for org.freedesktop.UPower.PowerProfiles
```

(Not sure if that's just because the name doesn't match `net.hadess.PowerProfiles`.)

jacobgkau:
The profiles do seem to be working normally again across the CLI and GUI in Pop!_OS now.

mmstick:
With the added dbus conf, I'm now able to change power profiles in Ubuntu 24.04's power settings page

jacobgkau:
Confirmed, the Ubuntu GUI is showing up (including in the top-right menu) with ae034d8, and it does work for switching profiles.

I did notice a bug with the top-right menu applet. While I'm able to switch power profiles successfully, the main button of the applet always says "Balanced" and does not reflect the actual active profile. See below:

[Screencast from 2024-04-30 14-18-53.webm](https://github.com/pop-os/system76-power/assets/7199422/b9442b88-5f0d-4d80-a13b-f8f037ecfd36)

If I replace `system76-power` with `power-profiles-daemon`, then the applet's preview/button does show the current profile:

[Screencast from 2024-04-30 14-22-39.webm](https://github.com/pop-os/system76-power/assets/7199422/c43adb76-96ea-4b7d-b7cc-12caec3f0d0c)

mmstick:
I'm not sure why gnome-shell is having difficulty here. Need to keep investigating. The power profiles controls in the settings panel is updating fine.

mmstick:
After extensive debugging, I've come to the conclusion that there is a regression affecting all versions of zbus 4.x.x that breaks property getters in GJS. What's unusual is that property getters work fine everywhere else. If I run `busctl --system get-property net.hadess.PowerProfiles /net/hadess/PowerProfiles net.hadess.PowerProfiles ActiveProfile`, I get the correct value. I was not able to find any issues when comparing the dbus interface in d-feet with the xml file in gnome-shell. And for whatever reason, property setting works fine from GJS. So it is only property getting that fails with `ActiveProfile` in GJS.

It's possible the zbus 4.2.0 has a fix, but I was not able to try it out because it's blocked by https://github.com/dbus2/zbus/issues/764

jackpot51 (dismissed):

jacobgkau:
On Ubuntu 24.04, the GUI and CLI are now working in conjunction as expected. I can adjust the power profile through the GUI or the CLI, and the GUI always reflects the correct current state. (The top-right drop-down and the Settings app are also kept in sync.)

On Pop!_OS 22.04, the duplicate menu does show up immediately after installing the updates, because GNOME Shell must be restarted to make the updated extension take effect. I don't see how we could get around that (but let me know if there is a way). That's not a serious issue, since the extra menu does partially work (it just isn't kept up-to-date) and goes away with the next restart.

On Pop!_OS 22.04, I do see that the Power Mode section still shows up in Settings after the system76-power update, and the gnome-shell-extension-system76-power menu is not updated if I switch the mode in the Settings app. Can we either fix that in https://github.com/pop-os/gnome-shell-extension-system76-power/pull/94 so the top-right menu gets updated if the Settings app is used to switch the mode, or else patch Settings to prevent that section from showing up after the update?

jacobgkau (approved):
Ubuntu 24.04 is still working as expected on 17a39f2.

Pop!_OS 22.04 is now also working as expected. The CLI and both GUI's are always in sync. (As previously mentioned, a duplicate menu will show up in the top-right drop-down until the next GNOME Shell restart after installing the updates, but the two menus are at least kept in sync now, too.)

<details>
<summary>Regression testing passed:</summary>

### CLI

    - [X] Command returns "not supported" message on a non-switchable system
- [X] Discrete graphics power state can be queried and set (compute mode)
    - Won't power back on after powering off (addw2). Not a regression.

### GNOME Shell

</details>

Tested with https://github.com/pop-os/gnome-shell-extension-system76-power/pull/94.

I found that these PRs fix https://github.com/pop-os/gnome-shell-extension-system76-power/issues/76.

jackpot51 (approved):

zeenix:
> It's possible the zbus 4.2.0 has a fix, but I was not able to try it out because it's blocked by [dbus2/zbus#764](https://github.com/dbus2/zbus/issues/764)

I did not but now both issues are fixed and I just rolled out zbus 4.2.1 that includes these fixes. I'd strongly recommend switching to zbus 4 as it's the only supported stable version. Thanks for reporting the issues btw.

PR #436, opened 2024-08-13
bonw15-b: TBT quirks, KBLED handling

jackpot51:
- Keep Thunderbolt controller active to work around display HPD issue
- Disable runtime power management for TBT xHCI
- Add handling for keyboard LED controls when using system76-dkms
- Check for `color_left` if `color` does not exist

### DIFF
--- a/src/daemon/mod.rs
+++ b/src/daemon/mod.rs
@@ -29,7 +29,7 @@ use crate::{
     hid_backlight,
     hotplug::{mux, Detect, HotPlugDetect},
     kernel_parameters::{KernelParameter, NmiWatchdog},
-    runtime_pm::runtime_pm_quirks,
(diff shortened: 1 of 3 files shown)


crawfxrd:
Will adding KBLED handling for system76-dkms break anything else? How do systems with proprietary firmware currently handling KBLED controls?

crawfxrd:
Does the color correction for bonw15 need to be applied to bonw15-b as well?

jackpot51:
> Does the color correction for bonw15 need to be applied to bonw15-b as well?

Yes

crawfxrd:
Colors look fine to me, and the bonw15 color correction causes white to look like light pink. Issue potentially fixed from bonw15? Probably just handled in firmware.

crawfxrd (approved):

jacobktm (approved):
Keyboard backlight hotkeys on bonw15-b are all working as expected now!

PR #437, opened 2024-08-14
fix: Match full_charge profile to firmware default values

jacobgkau:
This should make `system76-power charge-thresholds` report the "Full Charge" profile by default on Open Firmware laptops, instead of reporting a "Custom" profile by default.


mmstick (approved):

garrettjwilke (approved):
serw13 - tested all power profiles. behaving as expected

PR #439, opened 2024-08-31
add huawei support for charging thresholds

elevenhsoft:
Hello,
This simply checks if device have charge control thresholds ability. In this case, it should works with most of the huawei laptops. Tested on my Matebook and it works.


leviport (approved):
I tested charging thresholds on a lemp10 and an oryp12, and there are no regressions on either. I don't have a Huawei device to test with, so I'll defer to your testing on that front.

mmstick (approved):

elevenhsoft:
Thank you! I tested it again after `apt update` and yeah, it works after switching in battery applet :partying_face: 

```
~ ❯ cat /sys/devices/platform/huawei-wmi/charge_control_thresholds 
70 80
```

PR #450, opened 2025-03-25
Add bonw16

crawfxrd:


crawfxrd:
HDMI output doesn't appear to work at all on 24.04, so I haven't been able to check if this works.

The TBT quirks will probably also be needed since this is also using the TBT5 controller.

src/hotplug/mod.rs
```diff
@@ -166,6 +166,18 @@ impl HotPlugDetect {
                     ],
                 }),
             }),
+            "bonw16" => Ok(Self {
+                integrated: Integrated::Intel(Intel {
+                    sideband: Sideband::new(0xE000_0000)?,
```
jackpot51, line 171: Make sure to match this address to the one in coreboot-collector

jackpot51 (changes requested):

src/hotplug/mod.rs
crawfxrd: that's the new `0x9F_F000_0000` address for PCH-S?

jackpot51: Correct

jackpot51 (dismissed):

jackpot51 (approved):

leviport (approved):

PR #452, opened 2025-05-01
Add serw14

crawfxrd:
Same port as bonw16, so copied the config.


leviport (approved):
Prompt to switch to Hybrid only appears when I'm in Integrated graphics and plug a display into HDMI. Both USB-C ports drive displays in Integrated, and no prompt to switch modes appears.

jackpot51 (approved):

PR #460, opened 2025-08-10
fix: support for fedora (dracut + xorg path)

ozgb:
Fixed support for Fedora by choosing `dracut` over `initramfs-update`
if available, and checking the XOrg path

This may also bring support for other distros, but I haven't tested
those


mmstick (approved):

jacobgkau (approved):
<details>
<summary>Regression testing on Pop!_OS 22.04 passed:</summary>

### CLI

    - [X] Command returns "not supported" message on a non-switchable system
- [ ] Discrete graphics power state can be queried and set (compute mode)
    - Won't power back on after powering off (oryp6). Not a regression.

### GNOME Shell

</details>

Also works fine on 24.04, although graphics mode switching is deprecated and not exposed in the GUI in 24.04.

`dracut` is not the default for Ubuntu/Pop!_OS 22.04, but it is available to install from the Ubuntu repository (which removes `initramfs-tools` as a conflict). I found that system76-power was still able to switch graphics modes as expected after installing `dracut` (and that it indeed called `dracut` in the background, according to the system's logs).

I initially thought it might be smarter to have `update-initramfs` stay the default and only use `dracut` if it's not present. However, it seems Ubuntu has been in the process of switching its default tool to `dracut` for some time:

- Proposal in 24.10: https://discourse.ubuntu.com/t/please-try-out-dracut/48975/38 
- Official support in 25.04: https://discourse.ubuntu.com/t/dracut-will-be-supported-in-ubuntu-25-04/56556 

Considering that, the fact that a user somehow accidentally installing `dracut` will not cause any regressions per my smoke testing, and the fact that they can't be co-installed in Ubuntu anyway due to their packages conflicting, the way this is written seems fine.

PR #463, opened 2025-08-29
Remove switchable graphics code

Drakulix:
This drops all the graphics related code, so that users don't accidentally write config files using the `system76-power` cli tool, that will not work with COSMIC.

Special care has been taken to keep the bonw15-hack and not enable the fine-grained power-management on that machine. (Though this might only happen after first boot, but that isn't worse than the current solution.)

This should only be merged when `master_jammy` has been merged off to not break switching on 22.04 installs.


src/daemon/mod.rs
```diff
@@ -66,16 +66,18 @@ async fn signal_handling() {
 static PCI_RUNTIME_PM: AtomicBool = AtomicBool::new(false);
 
 // TODO: Whitelist system76 hardware that's known to work with this setting.
-pub(crate) fn pci_runtime_pm_support() -> bool { PCI_RUNTIME_PM.load(Ordering::SeqCst) }
+pub(crate) fn pci_runtime_pm_support() -> bool {
+    PCI_RUNTIME_PM.load(Ordering::SeqCst)
+}
```
crawfxrd, line 71: Looks like formatting was run that ignores `rustfmt.toml`.

Drakulix: I missed, that this repo wants to be formatted with rustfmt-nightly. Should be fixed.

jacobgkau:
This is working fine on a random hybrid system (addw5) in my testing so far. We may need to find a bonw15 around the lab to confirm its hack still works.

jacobgkau (commented):
Suspend/resume is working fine on a bonw15 with this branch.

On an addw5, if I first switch to integrated mode with `system76-power graphics integrated`, then install this branch, the NVIDIA dGPU is stuck off with no way to turn it back on. It seems like we should handle this somehow. We could:

- Leave only the `graphics hybrid` command for now, to allow people to get back to our supported config if they'd previously switched to something else.
- Branch `master_noble` and make this change only for `master`/Resolute, and make sure pop-upgrade resets them to Hybrid on the upgrade to 26.04.
    - Looking at pop-upgrade's code for resetting to the default mode on upgrade, it appears it depends on system76-power (via dbus) to do this right now.
    - system76-power also tells it (and the installer) to set some models to Integrated by default, based on if their GPU supports runtimepm or not. Not sure what the current status would be for users of laptops/GPUs that fell into that category before.

PR #468, opened 2025-09-23
Remove modeset, i2c_nvidia_gpu modprobe options

crawfxrd:
`modeset` is set by the driver package, so it is not necessary to explicitly set it.

`i2c_nvidia_gpu` is available as a quirk in system76-driver for models that have issues with suspend because of it.


jacobgkau:
Do we want this change for all branches, or only 24.04 like https://github.com/pop-os/system76-power/pull/463?

mmstick (approved):

PR #482, opened 2026-06-25
Limit `power/control` hack to coreboot firmware

crawfxrd:
Proprietary firmware should not require this hack.

Fixes issue where it was disabling runpm on addp6.
Underlying cause of why addp6 is not defaulting to hybrid is not addressed.

Cannot be dropped completely until coreboot firmware with `DGPU_PWRGD` fix is released.

Ref: system76/coreboot#274

### DIFF
--- a/src/graphics.rs
+++ b/src/graphics.rs
@@ -633,7 +633,12 @@ impl Graphics {
             log::info!("Enabling graphics power");
             self.bus.rescan().map_err(GraphicsDeviceError::Rescan)?;
 
-            sysfs_power_control(self.nvidia[0].id.clone(), self.get_vendor()?);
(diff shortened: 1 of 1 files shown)


jackpot51 (dismissed):

leviport (dismissed):

crawfxrd:
addp6 quirk may not be needed, after limiting `power/control` hack to models with coreboot firmware.
But would not explain why addp6 is not defaulting to hybrid graphics in the first place.

mmstick (dismissed):

jackpot51 (approved):

leviport (approved):

crawfxrd:
Man, fuck the ARM build. Someone merge this.

